PR #8, opened 2015-02-26
Rename licence.txt to license.txt

smith:


mike-thompson-day8:
Thanks!  Will be a part of upcoming v0.2.0 which is currently in develop

PR #14, opened 2015-02-28
fixed cond vs. condp typo in example code

btheado:


mike-thompson-day8:
Thanks!  Will be a part of upcoming v0.2.0 which is currently in develop.  Out soon.

mike-thompson-day8:
Thanks!! That change has now landed

PR #21, opened 2015-03-04
Slight wording tweaks

christianromney:
reactivity -> reactively (reads better in this context, imho)
while ever -> whenever


mike-thompson-day8:
Thanks.  Changes included in v0.3.0

PR #26, opened 2015-03-10
Update README.md

hura:


mike-thompson-day8:
Thanks. Fixed in v0.3.0

PR #30, opened 2015-03-14
Fix up the enrich docstring

sir-pinecone:
Fixed a typo and reworded things a bit to make it flow better, IMO.


mike-thompson-day8:
Thanks. Will be a part of v0.4.0

PR #31, opened 2015-03-15
Fix list formatting in README

sir-pinecone:


mike-thompson-day8:
Thanks.  Change will be a part of v0.4.0

PR #32, opened 2015-03-15
Fix a broken wiki link

ivanreese:


mike-thompson-day8:
Will be fixed as part of v0.4.0. Thanks!

PR #40, opened 2015-04-04
Fixed wiki link from README.md

timgilbert:


hura:
Get in line, this is the 3rd PR for this :)

mike-thompson-day8:
I am sorry about this.  The change has been made in develop weeks ago,  but I've just not pushed to master. 

I've been perpetually stuck in a mode of expecting to push 0.3.0 but then not doing it because I'm still no happy enough with the subscription tweaks I've made.  

I think I need to push a slightly less ambitious 0.3.0  which just fixes up a few things and hold subscriptions tweaks back to 0.4.0 so i can dwell on them a bit more.  I've also been very preoccupied trying to get re-com out the door. 

Anyway, will push 0.3.0 shortly (a very modest release) and then noone will ever have to fix this issue again.

timgilbert:
Oh, sorry - I should have checked the existing pull requests first. I'll close this one. I looked briefly for a way just to redirect the old page URL to the new one, but there doesn't seem to be a simple way to do it.

@mike-thompson-day8, thanks for re-frame, it's awesome. No reason to rush on a new release (though I understand wanting to get something out the door).

PR #44, opened 2015-04-17
Small typo fixes in comments

led:


mike-thompson-day8:
Many Thanks. You prompted me into spending a bit time tidying up that simple example.  It was done in haste originally.  So an improved version is now in develop branch and will be a part of v0.4.0

PR #57, opened 2015-05-05
Fix 'where do I go next' TOC link in README

sir-pinecone:
My editor removes trailing spaces when I save so that explains the extra changes.


mike-thompson-day8:
Thanks!

PR #61, opened 2015-05-18
Readme clarifications

ricardojmendez:
Hi,

Here's a couple of fixes for minor README issues on things that had me wondering on my first read (for instance, where did I set up which app-db to use).


mike-thompson-day8:
Thanks

ricardojmendez:
Thank you for making re-frame actually - it made things an order of magnitude cleaner.

yatesco:
+1

PR #76, opened 2015-07-11
Simplify implementation of Middleware Compose

estsauver:
The implementation of middleware compose is duplicating behavior already
in compose.

`(apply comp [])` returns identity.


danielcompton:
Nice!

mike-thompson-day8:
Thanks!

Issue #82, opened 2015-07-14
Readme.md inaccuracy regarding Elm


bstro:
In the blog post at http://elm-lang.org/blog/announce/0.14, Elm evidently no longer uses `lift`, instead calling it `map`. re-frame's README.md states `While the mechanics are different, reaction has the intent of lift in Elm and defc= in Hoplon.` This might be confusing for certain readers.

mike-thompson-day8:
Thanks for the correction.  Will roll out that tweak shortly.

PR #83, opened 2015-07-15
Change reference to Elm syntax from 'list' to 'map'

Umedoc:
https://github.com/Day8/re-frame/issues/82


mike-thompson-day8:
Thanks for the correction!! 

I'm part the way through making a couple of changes to the REAME on the develop branch,  so I'll manually make this tweak in what I'm doing.  Should be rolled out soon.

Issue #94, opened 2015-07-25
TodoMVC example looks bad on Firefox, Clear Completed button doesn't work


thenonameguy:
Tested with Firefox 39.0 on OSX 10.9.5

Chrome:
<img width="598" alt="screen shot 2015-07-25 at 3 05 45 pm" src="https://cloud.githubusercontent.com/assets/2217181/8889549/dceed066-32de-11e5-9170-2b6f5b7aa191.png">

Firefox:
<img width="593" alt="screen shot 2015-07-25 at 3 06 32 pm" src="https://cloud.githubusercontent.com/assets/2217181/8889550/e44e6880-32de-11e5-9455-48e8c245c9f3.png">

Clicking Clear Completed doesn't trigger the dispatch, nor does rewriting the `:on-click` to `(js/alert 'test')`. Also rewriting the text from 'Clear completed' to anything doesn't get reloaded with figwheel, but (oddly) every other part of the page gets rewritten correctly/works.

Very odd. Couldn't find a solution. Halp.

kamn:
Looking into it there is a css class that is setting the content to be 'Clear completed'. It completely ignores the cljs view code. I am working on a pull request for fixing it.

kamn:
Also in case anyone else comes along and wonders why Firefox has checkboxes it boils down to [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=605985). You can't easily restyle checkboxes like you can in webkit.There are some solutions in the comments if someone wants to try though.

thenonameguy:
Thanks @kamn!

mike-thompson-day8:
When doing todomvc, you have to use the offical CSS file from https://github.com/tastejs

I seem to remember using something like this: 
https://github.com/tastejs/todomvc/blob/master/examples/backbone/node_modules/todomvc-app-css/index.css

Have there been changes to cover the firefox issues?

mike-thompson-day8:
I'm going  to close this issue.  As far as I can tell, we're using a pretty up to date CSS file from the official todomvc site.  If that is wrong, please provide a PR with the correct  CSS.

PR #95, opened 2015-07-25
Fixed issue with css overwriting the cljs views for "Clear completed"

kamn:
See #94.

Fixes the issue with the "Clear completed" being rendered from the css and not the cljs code and also correctly fires the event to clear todos. I think the css was a little weird for Firefox and didn't allow the event to be fired because of `visibility:none`?

Though this does not solve the issue with the view of the check-boxes in Firefox.

I wonder if this css was added for moble? Also in the cljs code it wants to render like...
"Clear completed #" I just commented the # part out. Not sure if that is wanted anymore or not.

### DIFF
--- a/examples/todomvc/resources/public/todos.css
+++ b/examples/todomvc/resources/public/todos.css
@@ -323,19 +323,17 @@ html #clear-completed:active {
     line-height: 20px;
     text-decoration: none;
     cursor: pointer;
-    visibility: hidden;
     position: relative;
 }
 
 #clear-completed::after {
     visibility: visible;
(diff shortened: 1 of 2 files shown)


thenonameguy:
@mike-thompson-day8 you closed #94 with a comment stating that you are looking for a PR that fixes the problem mentioned in that issue. This fixed it for me, can you take a look at it?

kamn:
You need to use the official css which is the issue. I will check later today if the css is updated or not but I think mike has already checked. Until then I am going to close this.

PR #101, opened 2015-07-27
DRY logger assignment

thenonameguy:
We could use the fact that the binding names match the keys to even simplify further, but I'll leave that up for discussion.

### DIFF
--- a/src/re_frame/utils.cljs
+++ b/src/re_frame/utils.cljs
@@ -30,11 +30,15 @@
   (swap! loggers merge new-loggers))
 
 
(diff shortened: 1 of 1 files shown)


danielcompton:
Hey @thenonameguy, I really appreciate this PR (and your others). I think we're going to decline this for now as there isn't a large benefit to the changes, and we lose Cursive's code analysis if we use `def` instead of `defn`. I understand it's not the DRYest, but this section of code is also pretty static, so bugs in this area seem pretty unlikely, and every change carries risk. So it's a no for now, but we do appreciate your help, and we may have a change of heart on this in the future.

Thanks!

PR #103, opened 2015-08-12
Updated the CSS from todomvc and change the views to match

kamn:
In reference to #94 (and a fix for it).

I pulled the latest version of the Todomvc css from [here] (https://github.com/tastejs/todomvc-app-css). The big change they made was from ids to classes so I had to change the views also. There is one functionality change in that I removed the count of issues since the current examples don't do that.

### DIFF
--- a/examples/todomvc/resources/public/todos.css
+++ b/examples/todomvc/resources/public/todos.css
@@ -1,364 +1,348 @@
 html,
 body {
-    margin: 0;
-    padding: 0;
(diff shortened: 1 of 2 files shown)


thenonameguy:
LGTM

kamn:
Ah sorry about that. I will fix it later today.

danielcompton:
I think so, it doesn't seem possible to change it once you've opened the PR.

PR #107, opened 2015-08-16
decouple re-frame and reagent+core.async

darwin:
This rewrite was motivated by https://github.com/Day8/re-frame/pull/106.

In essence re-frame provides a transducer: state, event -> state.
This transducer is able to apply algorithmic transformations on series of events
and build final state up. This allows complete decoupling of [re-frame handlers,
subscriptions and middleware logic] from [mechanisms how events are queued,
processed and how are their effects applied to app-db]. This "bare re-frame" is
implemented in frame.cljs.

For maximal flexibility bare re-frame must be pure and have no knowledge
of app-db, reagent/ratom and core.async. It must be possible to create multiple
independent instances of bare re-frame.

For convenience scaffold.cljs is re-implementing original re-frame functionality
of v0.4.1 using those new bare primitives. This is exposed as public api via
core.cljs.

App developer opts-in using default implementation by requiring re-frame.core.
In this scenario re-frame creates a single app-db (reagent's ratom) and
runs single event loop where events are dispatched and processed asynchronously 
using core.async channel. A single re-frame instance is created and kept in
re-frame.core/app-frame atom. Plus developer gets access to the original
API to manipulate state of this single re-frame instance.

App developer is also free to pick/implement other means of employing bare
re-frame by not requiring re-frame.core and building her own scaffold
around bare re-frame primitives.


darwin:
This PR will need more work to achieve 100% compatibility. 

I wrote a few tests. I plan to write more. Also made sure simpleexample works.

Open issues are:
1) use of legacy-subscribe
2) missing pure handler
3) whole undo functionality is missing

project.clj
```diff
@@ -1,43 +1,43 @@
-(defproject re-frame "0.4.1"
-  :description  "A Clojurescript MVC-like Framework For Writing SPAs Using Reagent."
-  :url          "https://github.com/Day8/re-frame.git"
-  :license      {:name "MIT"}
-  :dependencies [[org.clojure/clojure "1.6.0"]
-                 [org.clojure/clojurescript "0.0-3211"]
+(defproject re-frame "0.4.2-TRANSDUCERS"
+  :description "A Clojurescript MVC-like Framework For Writing SPAs Using Reagent."
+  :url "https://github.com/Day8/re_frame.git"
```
danielcompton: why change to `_`?

darwin: ah, that was unintentional, I was renaming a folder under tests directory and intellij was too smart, I guess. Will change it back.

src/re_frame/frame.cljs
```diff
@@ -0,0 +1,119 @@
...
+  (:require [re-frame.utils :refer [get-event-id get-subscription-id simple-inflection frame-summary-description]]
+            [re-frame.logging :as logging :refer [log warn error]]))
+
+; re-frame meat reimplemented in terms of pure functions (with help of transducers)
+
+; see http://clojure.org/transducers[1]
+(defn get-frame-transducer
+  "Returns a transducer: state, event -> state.
+This transducer resolves event-id, selects matching handler and calls it with old db state to produce a new db state.
+Tranducer must have no knowledge of underlying app-db-atom, reagent, core.async or anything else out there.
```
thenonameguy: Typo in `Tranducer`.

thenonameguy:
Looks great, but could you clean up the logging code a bit with something like my #101 PR?

darwin:
Ok, now I think the code is in pretty good shape. Both simpleexample and todomvc work without source code changes on their side.

Outstanding problems are IMO worth breaking semantics/API for upcoming release.

Note: by "Frame" I mean code in frame.js, which is bare re-frame, stripped from reagent/core.async knowledge.

#### 1) register-sub should not promise passing app-db
The signature of subscription handler is [app-db-atom subscribe-param]. First, it forces Frame to know about this opaque app-db-atom thing and pass it when someone creates new subscription. Second, it is not general enough. One can imagine completely different subscription systems, which do not depend on reagent's ratoms/reactions. And yet, Frame could be used with them if proper scaffolding is provided.

Current solution: I have implemented backward-compatible register-sub which pipes app-db-atom through, but it had to be pushed into scaffold: 
https://github.com/darwin/re-frame/blob/9b318895b66749b577ed3e8e6d8e94e4f0743a0e/src/re_frame/scaffold.cljs#L39-L46

My proposal is to change contract of register-sub. The signature handler passed will have [subscribe-param] signature. If needed, app-db can be passed into subscription as regular param in the vector. Or handler can reference it globally. Or for convenience user can create her own subscribe method, which will prepend common parameter like that automatically.

Secondary question is why having this subscription subsystem as part of Frame's state anyway. Maybe it would make more sense to have it as a separate state object independent from event handling.

#### 2) pure middleware makes no sense anymore

Handlers (composed of middlewares) should have no knowledge about outside world. They should not care how are events delivered and how they are applied to app-db. Handlers should be only responsible for producing new db state given previous state and incoming event.

Current solution: I have included pure middleware for compatibility reasons, but it is not used by default re-frame implementation anymore and it will log a warning if really called by someone.

#### 3) undo functionality was removed and needs to be redesigned IMO

I feel like undo/redo functionality should not be middleware or register its own handlers and subscriptions in re-frame.

I think it should be responsibility of the event processor (who knows about app-db and rules of event queueing) to provide simple functional API for it and keep the undo/redo state. In default case this should be implemented on scaffold-level. It is up to app developer if she wants to expose calls to this api via events.

src/re_frame/middleware.cljs
```diff
@@ -189,13 +164,13 @@
       (fn after-handler
         [db v]
         (let [new-db (handler db v)]
-          (f new-db v)   ;; call f for side effects
+          (f new-db v)                                      ;; call f for side effects
```
thenonameguy: Why so much whitespace?

darwin: Sorry, I have reformatted it with Intellij/Cursive, it is indenting comment like that. 

Feel free to edit it. I won't be touching it that much.

darwin:
I'm happy to announce a success story with Plastic. I was able to switch to this fork and use re-frame as a library:
https://github.com/darwin/plastic/commit/1e87f07b8b929d8d17e21dff1d62bb6471e5a477

My custom scaffolds for main/worker thread routers are less 100 LOC:
https://github.com/darwin/plastic/blob/master/cljs/src/main/plastic/main/frame.cljs
https://github.com/darwin/plastic/blob/master/cljs/src/worker/plastic/worker/frame.cljs

src/re_frame/frame.cljs
```diff
@@ -0,0 +1,175 @@
...
+                      ([db-states new-db]                   ; in each step
+                       (conj db-states new-db)))            ; add new-db state to the vector
+        xform (get-frame-transducer frame last)]
+    (transduce xform reducing-fn [init-db] events)))
+
+(defn process-event-on-atom [frame db-atom event]
+  (let [old-db @db-atom
+        new-db (process-event frame old-db event)]
+    (if-not (identical? old-db new-db)
+      (reset! db-atom new-db))))
```
thenonameguy: This pattern:
```clojure
(if-not (identitcal? old-db new-db)
  (reset! db-atom new-db))
```
is appearing in quite a few places, it might be better refactor it out to a separate fn:
```clojure
(defn reset-if-changed! [db-atom old-db new-db] (...))
```

thenonameguy:
LGTM

thenonameguy:
@danielcompton, @mike-thompson-day8  this PR has been sitting idle for a while, but would make great improvements to the code base. Can we get an estimate when this will be reviewed/merged? Having multiple re-frame instances on a single page would be really beneficial to my use case.

danielcompton:
@thenonameguy honestly these changes are massive (but really interesting), and I'm going to need some time to chew over them. Mike and I are both in a pretty busy phase of work at the moment and speaking for myself I don't have the time or mental capacity to review it right now. Ballpark estimate, I doubt I'd be able to look at it for a few weeks *at least*, and I suspect Mike is in a similar boat to me.

If you're keen to work with this in the meantime, you could always take this patch and run a forked version for a while. Your feedback on how it feels from an API perspective would also be helpful here.

I know that's not perhaps not the answer you were looking for, but I don't have a lot of time or energy to give right now.

danielcompton:
Also, Day8 have what I presume to be the largest production Re-Frame codebase, and we don't want to break our own apps or get stuck on an old version without having good mitigations for the points that @darwin has mentioned above.

thenonameguy:
Sure, thanks for the response. Then I'll just start using a fork as you mentioned.

darwin:
For people interested in this version. I'm going maintain the fork here:
https://github.com/binaryage/pure-frame

Just released v0.1.0 as `binaryage/pure-frame`

dijonkitchen:
Any updates or should it be closed?

PR #115, opened 2015-09-17
implements? is a macro - can't use with partial

mccraigmccraig:
implements? can't be partially applied - so wrap it in a function instead

this PR isn't against the develop branch because #108 isn't merged there yet


danielcompton:
Hey thanks for catching this, I really appreciate it. I've fixed this on the `dynamic-subscriptions` branch directly (I just rescued the previous commit I had which took this approach).

Thanks!

PR #120, opened 2015-09-27
Use yield instead of (timeout 0)

danielcompton:
`(timeout 0)` resolves to using `js/setTimeout` which will actually take 4+ msecs in browsers. This is an eternity. The nextTick approach was proposed by Patrick O'Brien and implemented here.

Docs on nextTick from Closure library:

> Fires the provided callbacks as soon as possible after the current JS execution context. setTimeout(…, 0) takes at least 4ms when called from within another setTimeout(…, 0) for legacy reasons.

> This will not schedule the callback as a microtask (i.e. a task that can preempt user input or networking callbacks). It is meant to emulate what setTimeout(_, 0) would do if it were not throttled. If you desire
microtask behavior, use goog.Promise instead.

Unfortunately when I use this pull request against the Todo MVC example, and an internal Day8 project, I get [this error](https://gist.github.com/danielcompton/e4859424946a8ac6fae0). @pkobrien do you have any thoughts or ideas on this?


pkobrien:
I wish I had something to contribute. Seems like a very strange error. I'll be curious to know what you find out.

danielcompton:
Closing this PR as it is against master, not develop, continued on #121.

PR #124, opened 2015-10-16
Cache subscriptions

danielcompton:
Currently every time someone subscribes, they will create a new Reaction. The same thing happens when a subscription creates subscriptions in it's own creation. The net result of this is that it is possible to waste a lot of CPU time recalculating the same Reactions.

This patch adds a subscription cache. When a subscription is requested, we check the cache. If it exists in cache then we return the cached Reaction, if not then we create the subscription, add it to the cache,
and return it.

The tricky part is cache invalidation (isn't it always?). I'm not 100% sure how to remove Reactions that are no longer required.

One option is to return a Reaction like

```
(let [sub (reaction (handler-fn app-db v) :on-dispose #(clean-up)]
```

however that means we have to pay for two Reactions, the `handler-fn`, and the extra reaction to dispose. I'm not sure whether the cost of two Reactions outweighs the benefits of caching subscriptions. Performance testing would be helpful here.

At first I thought caching dynamic subscriptions was going to be tricky. But it actually works out fine. For example in:

```clj
(register-sub
  :todo-dynamic
  (fn todo-dynamic [_ _ [active-list]]
    (let [q (q/get-query active-list)]
      q)))

(register-sub
  :todos
  (fn todos [db _]
    (let [active-list (subscribe [:active-list])
          todos       (subscribe [:todo-dynamic] [active-list])]
      (make-reaction (fn todo-vals [] (update @todos :result #(vals (:list %))))))))
```

the dynamic parameter `active-list` passed to `:todo-dynamic` will be cached. This means it will always be the same, so cache will work correctly.

More tests are needed here to make sure all of the caching behaviour is correct.

### DIFF
--- a/src/re_frame/subs.cljs
+++ b/src/re_frame/subs.cljs
@@ -7,6 +7,8 @@
 
 ;; maps from handler-id to handler-fn
 (def ^:private key->fn (atom {}))
+(def ^:private v->subscription (atom {}))
+(def ^:private dynv->subscription (atom {}))
 
 
 (defn clear-handlers!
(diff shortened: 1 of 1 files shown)


mike-thompson-day8:
Related:  https://github.com/reagent-project/reagent/issues/201

danielcompton:
This is fixed in 3d4e8cde8ff29c640fc7dac0292975e1b019e95e. It gets around the double reaction disposing problem by using a new feature in Reagent 0.6.0 to add an `on-dispose` to an existing reaction.

PR #128, opened 2015-11-01
Small path middleware change

nowprovision:
update-in can take extra args


mike-thompson-day8:
We only take patches against the develop branch.  
Given how tiny the change is, I'll just manually do it across on develop.  
**But thanks for the tweak. **

Issue #125, opened 2015-10-17
Testing for 0.5.0 release


danielcompton:
Hi folks

We've released https://github.com/Day8/re-frame/releases/tag/v0.5.0-alpha1. Try it out in your apps and let us know what you think. There are two changes that need particular attention:

* Dynamic subscriptions - https://github.com/Day8/re-frame/pull/108. Try them out and let us know if the API feels right for this. I'm pretty sure that we don't have any memory leaks in our implementation of dynamic subscriptions, but let us know if something seems funky. More docs for this will be forthcoming
* Switching from `(core.async/timeout 0)` to `(yield)` in the router dispatch loop - https://github.com/Day8/re-frame/pull/121. We need to give control back to the browser while we're processing for it to do Browser Things. We previously used `(timeout 0)` to do this. The time taken for the  `(timeout 0)` channel to close is browser specific, but in general takes at least 4 ms. This meant we couldn't process more than 250 events/second. The new `yield` implementation, helpfully shared by https://github.com/pkobrien is much faster. Again, we think this is implemented correctly, but let us know if you see anything funky.

Thanks!

scgilardi:
summary: re: (yield) change: I see a nice improvement in my app's performance, but perhaps we can get an even better balance between the needs of the browser and the needs of the router.

I'm writing an app that uses svg to allow the user to draw items onto a drawing surface. Currently I have one entry in the db when a new item is being created by a mouse drag and a separate entry for the list of items that already exist. In the "mouse up" handler for a creation drag, I dispatch two re-frame events: :end-drag, and :create-item. Before the yield change this gave a significant visual glitch as the drawing surface was redrawn on each yield: first without the new item (after :end-drag was handled) and then again with the new item in the existing items list (after :create-item was handled).

In this case, the effect of the yielding to the browser before handling each event was a visual flicker ("there" "not there" "there") during item creation. The duration of "not there" is greatly reduced in my setup after the change to use (yield) instead of waiting for (timeout 0). Sweet!

It does appear though that even with the new faster yield, software down the stack from re-frame (reagent/react/chrome) is rendering my svg every time the router loop yields. That seems like too much deference to UI handling over router event loop handling.

I see in the comments that the purpose of the yield to make sure the browser gets some time in case the handler hogs the cpu for a long time. I suspect many handlers are extremely fast (manipulations of the db atom) and it seems to me we should try to handle as many as we can in some reasonably small time slice rather than yielding before every event.

[...]

PR #129, opened 2015-11-04
proposed updates to new router, re: issue #125

scgilardi:
captures the bug fix and terminology suggestion from my comment: https://github.com/Day8/re-frame/issues/125#issuecomment-153569317


mike-thompson-day8:
Thanks!

PR #132, opened 2015-11-07
simplify the router

scgilardi:
- same behavior as before, but simplified by leveraging the fact that
  there is only one important state boolean: `(empty? queue)`

I was intrigued by the version 0.5.0 router's state machine and continued to work on making it simpler while having the same behavior. I focused on how the states actually affected behavior. I got to an intermediate stage with three states: `:idle`, `:waiting`, and `:running`, but then realized that `:waiting` and `:running` had disjoint sets of transitions and might as well be the same state. I was then left with only `:idle` and `:processing`. From there, It became clear that `:idle` is true if and only if the queue is empty.

This led to more simplification as the fsm-state variable could be removed and effectively replaced by a call to `(empty? queue)` whenever it needs to be consulted.

I think the set of methods in IEventQueue is now minimal and that the code is compact and easy to follow. Each function that remains is either called from more than one place or is called from a callback.  I can see a benefit in having an explicit state variable to enforce only the expected transitions, but the simplicity gained by removing it leads me to offer this simpler router implementation for your consideration.

In the comments, I noted this invariant to describe the state machine nature of the implementation proposed here:

```
;; EventQueue invariant:
;; - queue processing is in progress if and only if the queue is not empty
;;   - where "in progress" means either:
;;     - running now, or
;;     - waiting for a callback to run
;; Queue processing starts when an event is added to an empty queue, and
;; stops when the queue has been drained.
```

One thing that occurred to me as I considered what it takes to keep the machine running: we're relying on the callbacks to be absolutely reliable. If we miss a single one, the machine is likely to shut down and never recover. Is the assumption of perfect reliability close enough to reality to warrant keeping the machine as simple as it is? If this were a distributed system we'd some more robustness in the face of the possibility of missing callbacks.

One option would be to keep track of the time since the most recent event was processed and add another way for a "run-queue" callback to be scheduled. I'm thinking of a policy like:

- if you're about to enqueue an event and the queue is not empty, but no events have been processed in 2 seconds, schedule a `run-queue` callback to try to heal from what was likely a missed callback.

Given that this is all ultimately single-threaded, the cost of extra `run-queue` callbacks is approximately zero. In contrast, the cost of the machine shutting down because of a single missed `run-queue` callback is very large. I'm curious if you have any opinion about whether the possibility of missing a callback is even worth considering in the browser environment.

### DIFF
--- a/src/re_frame/router.cljs
+++ b/src/re_frame/router.cljs
@@ -23,7 +23,7 @@
 ;;
 ;; The original implementation of this router loop used core.async. It was
 ;; fairly simple, and it mostly worked, but it did not give enough
-;; control. So now we hand-roll our own, finite-state-machine and all.
+;; control. So now we hand-roll our own.
 ;;
 ;; The strategy is this:
 ;;   - maintain a queue of `dispatched` events.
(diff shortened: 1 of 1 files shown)


mike-thompson-day8:
@scgilardi  I actually started this process with a very terse version of the router/scheduler, which worked, but I threw it away because it didn't explicitly model the process as a FSM.  The subtleties of transitions were hidden behind an `if` here and a *not* doing a transition there.  so, it was terse as hell, but I worried that changes in the future might be tricky (not that I could imagine any future changes).  

Similarly, I feel your solution is lovely and terse but at the expense of it being harder for a new pair of eyes to understanding the subtleties later. The essence of the problem has been optimised away.

In effect, if I was sure there would never be a future changes, or if I had performance problems,  I'd be happy with either of our more terse versions. But I like the explicit, more verbose, slower version as future proofing. 

Of course, if there was a simpler version of the FSM then that would be good.  I didn't try too hard to make it minimal.  I wanted it easy, more than minimal. 

Regarding your question about "callbacks being reliable",  I think that's a reasonable assumption.  If callbacks were going missing,  I feel we'd be in a lot of pain all around.

scgilardi:
I understand. Thanks for taking a look!

PR #133, opened 2015-11-18
Added a space to make list in README.md appear

kamn:
Just added a space to make a list. I do have a questions about lists in general. Some times a list item is capitalized but more often it is not even if the point is a sentence. Would you mind if I got though and apply what's recommend in [this english.stackoverflow](http://english.stackexchange.com/a/55049) response?


mike-thompson-day8:
Thanks!

PR #138, opened 2015-12-10
Switch to using CircleCI and automated testing with Karma

danielcompton:
This is ready and has docs. I'm ready to merge it, unless @mike-thompson-day8 wants to review it first.


honzabrecka:
I just want to let you know that `karma start --single-run=false` does not work with larger code bases, because cljs compiler (or google closure compiler, I don't know) writes compiled files in chunks. So even though you increase the value of `autoWatchBatchDelay`, karma may start with incomplete output.

danielcompton:
Hey, thanks for the heads up. That does explain some of the errors I saw on our internal projects running with cljsbuild auto. I don't think it would be an issue here though, because we're doing a full cljsbuild before we even start up karma. Is that correct?

honzabrecka:
Yes, it's correct.

PR #147, opened 2016-01-22
Correct doc reference to handlers/register-base

mbertheau:


danielcompton:
Thanks for this. I've merged it into `develop`. It's not a massive deal, but if you can, open PR's against `develop`, not `master`. I do this too myself though, so don't worry too much about it. Thanks!

PR #149, opened 2016-01-27
Fix behavior with large undo lists

SerCeMan:
Hello!

It seems like undo doesn't work with large undo lists, because you always grab first 50 elements of a vector. I have tried "undo" in my application and have got wrong results.

Please, fix me if I didn't understand the idea correctly.

### DIFF
--- a/src/re_frame/undo.cljs
+++ b/src/re_frame/undo.cljs
@@ -53,10 +53,10 @@
   "Stores the value currently in app-db, so the user can later undo"
   [explanation]
   (clear-redos!)
-  (reset! undo-list (vec (take
+  (reset! undo-list (vec (take-last
                            @max-undos
                            (conj @undo-list @app-db))))
-  (reset! undo-explain-list (vec (take
+  (reset! undo-explain-list (vec (take-last
                                    @max-undos
                                    (conj @undo-explain-list @app-explain))))
   (reset! app-explain explanation))


SerCeMan:
Fix works for me

SerCeMan:
And also, the possibility to inject middleware to every handler without encapsulating register-handler will be a great feature. (I need possibility to inject my undo middleware to every handler in dev mode)

danielcompton:
Hi @SerCeMan, thanks for this! re-frame isn't big on tests in general, but this strikes me as a case where some tests could be helpful. Could you add some that fail on the old behaviour and pass on the new?

As far as injecting handlers into lots of middleware goes, you could make a `(def dev-middleware [...]` to use. Would that work?

mike-thompson-day8:
@SerCeMan   We started off thinking we needed `undoable` on all events, and then we realised we didn't.   

We have this sort of a case:  the user clicks on a button, generating a `dispatch`,  and the event handler initiates a server connection,  whose `on-success` handler initiates a further `dispatch`, and the associated event hander processes that response.  And ... you don't want `undoable` on all those dispatches.  Just the first, user initiated one, clicking on the button.  otherwise your users can step back into intermediate states, like twirly-thing-showing-while-waiting-for-server-response.

danielcompton:
Don't worry about the tests @SerCeMan, I'm working on them now.

danielcompton:
Thanks for this! I've merged your changes in https://github.com/Day8/re-frame/commit/bb78981e1d2cb906b82c08c9f942763c4dbb762f.

SerCeMan:
Thanks! About undoable, I use it not only for undo/redo behaviour but also for showing the difference between versions in dev mode, it very convenient for me.

danielcompton:
You might also find https://github.com/Day8/re-frame/wiki/Debugging convenient for that, it lets you trace individual function calls

PR #153, opened 2016-02-15
Fix docstring references to old fns

cldwalker:
Hi. I was reading the source and I thought these references were stale. Thanks for the well commented library!


mike-thompson-day8:
Thanks!

PR #156, opened 2016-02-29
Update terms used in Elm to latest changes

yang-wei:
Since Elm 0.14, there is no more `lift`. Fixed it to `map`.

> The term lift is dead. It makes me a bit sad for my thesis, but I think it will help a lot of people get started with signals more quickly. The new term is map, and the goal is to build on the intuition people have from working with lists.

http://elm-lang.org/blog/announce/0.14


mike-thompson-day8:
Thanks!

PR #167, opened 2016-05-04
middleware/debug: Log a single line if app-db not changed by a handler

### LINKED ISSUE #166: Change middleware/debug to only create a grouped diff when there is actually a diff
It would be helpful for scanning the console if middleware/debug had a different log message when there was no diff. For example I think:

screenshot of google chrome 5-05-16 10-59-15 am

would be more helpful if it was something like

```
clojure.data/diff for: [:my-event "with-val"]: no changes
```

danielcompton:
middleware/debug logs a group to the console even if there has been no change to app-db. This makes it harder to tell if you've made a mistake in refactoring, as the same log message will be shown whether or not app-db is changing.  

This change will log a single line if app-db is not changed by a handler

I did a quick performance check to make sure that my destructuring of the vector returned by `diff` didn't have a performance impact. There was no measurable difference.

Fixes #166

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,18 +1,23 @@
+## Unreleased
+
+Improvements
+  - `debug` middleware logs a single log line instead of a group if there is no difference in app-db between before and after running the handler.
+
 ## 0.7.0  (2016-03-14)
 
 Breaking:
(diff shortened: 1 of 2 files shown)


mike-thompson-day8:
Thanks!!  Looks useful. .

PR #177, opened 2016-07-14
Insignificant typo correction in console output

snowell:
Might bother only me, and I need to move to the new `def-sub` anyway, but hey!


danielcompton:
Thanks!

Issue #164, opened 2016-04-21
Preserve stack frames to show where event came from


danielcompton:
The re-frame router uses an FSM to process events, and puts events in a queue to be processed. This has many advantages, but one of the disadvantages is that you lose the context for where the event was originally called from. For example:

<img width="450" alt="screenshot of google chrome 21-04-16 12-54-32 pm" src="https://cloud.githubusercontent.com/assets/811954/14694808/4d91f418-07c0-11e6-8806-48e43d024f15.png">

Here we can see the stack trace that led to the error, but it stops at the `onmessage` trigger, and we don't know which part of our code called the `dispatch` that led to it. It would be great if we could keep stack frames from the dispatch call around.

https://datahero.com/blog/2014/05/22/node-js-preserving-data-across-async-callbacks/
https://bugs.chromium.org/p/chromium/issues/detail?id=332624

mike-thompson-day8:
As a solution, I'm wondering if you could write your own function called, say, `emit` and use it in place of `dispatch` everywhere.  `emit` would look something like this:
```
(defn emit 
   [event-vec]
   (when ^boolean js/goog.DEBUG console.trace((str v))
   (re-frame/dispatch event-vec))
```

Or go to the effort of writing `emit` as a macro.

danielcompton:
Here's another example where the stack trace tells you almost nothing:

<img width="797" alt="screenshot of google chrome 5-05-16 12-38-56 am" src="https://cloud.githubusercontent.com/assets/811954/15014392/d5407490-1259-11e6-85d0-be074a8633f4.png">

mike-thompson-day8:
Via slack discussion, @darwin introduced me to a sneaky idea which I'll quickly note here so it doesn't get lost:

1. Create a macro called say `dispatch2`
2. Within this macro, create an anonymous function which wraps `console.log`
3. Attach this anonymous function to the event's metadata, perhaps using the `:logger` key. 
4. When later it comes time to report what event is being processed, use this `:logger` obtained from the event itself. 

As a result the log message produced will appear to come from the place where the anonymous function itself was defined, which will be the point of the `dispatch` itself.

mike-thompson-day8:
So this ticket is about remembering where a `dispatch` happens.  If there are problems in the handling of that event, it is useful to know here it originally came from.  The async nature of event handling means stack traces are useless because they report that the event came from the scheduling software (and not the dispatch point). 

[...]

PR #179, opened 2016-07-19
Stack dispatch

stumitchell:
Proposed fix for https://github.com/Day8/re-frame/issues/164


mike-thompson-day8:
I merged this  manually.

PR #183, opened 2016-07-20
First pass at converting to CLJC for JVM testing

danielcompton:
Note that the goal of using CLJC here is to enable writing and running
`deftest`s in JVM Clojure to test your (CLJS) re-frame app.  Actually
running a re-frame app as a meaningful thing on the JVM (other than for
tests) is explicitly *not* a goal.

- All files renamed *.cljs -> *.cljc
- New `re-frame.interop` ns for stuff that's different between CLJS and
  JVM Clojure, including Reagent interop.
- Other CLJS incompatibilities (eg `catch :default` and `deftype`'s
  `^:mutable` vs `^:volatile-mutable`) are handled with reader
  conditionals in-place.


src/re_frame/interop.clj
```diff
@@ -0,0 +1,43 @@
+(ns re-frame.interop)
+
+(defn next-tick [f]
+  (.start (Thread. f)))
```
danielcompton: We should use a thread pool here, or just call `f` directly?

src/re_frame/interop.clj
```diff
@@ -0,0 +1,43 @@
...
+
+(def empty-queue clojure.lang.PersistentQueue/EMPTY)
+
+(def after-render next-tick)
+
+(def debug-enabled? true)
+
+(defn ratom [x]
+  (let [result (atom x)]
+    (alter-meta! result assoc ::ratom? true)
```
danielcompton: Instead of altering metadata, can we create a new `IReactiveAtom` protocol, and extend it to the atom?

src/re_frame/interop.clj
```diff
@@ -0,0 +1,43 @@
...
+    result))
+
+(defn ratom? [maybe-ratom]
+  (and (satisfies? clojure.lang.IAtom maybe-ratom)
```
danielcompton: If we had an `IReactiveAtom` protocol, we could simplify this to just `(satisfies? IReactiveAtom maybe-ratom)`

src/re_frame/interop.clj
```diff
@@ -0,0 +1,43 @@
...
+(def after-render next-tick)
+
+(def debug-enabled? true)
+
+(defn ratom [x]
+  (let [result (atom x)]
+    (alter-meta! result assoc ::ratom? true)
+    result))
+
+(defn ratom? [maybe-ratom]
```
danielcompton: Using `x` would probably be more idiomatic here.

src/re_frame/interop.clj
```diff
@@ -0,0 +1,43 @@
...
+(defn make-reaction
+  "On JVM Clojure, return an atom-like thing which invokes the given function on
+  every deref. That is, `make-reaction` here provides precisely none of the
+  benefits of `reagent.ratom/make-reaction` (which only invokes its function if
+  the reactions that the function derefs have changed value). But so long as `f`
+  only depends on other reactions (which also behave themselves), the only
+  difference is one of efficiency. That is, your tests should see no difference
+  other than that they do redundant work."
+  [f]
+  (reify clojure.lang.IDeref
```
danielcompton: Should we make a reaction here that implements the Reagent protocols? That way we are closer to the impl and avoid risk of diverging behaviour.

src/re_frame/interop.clj
```diff
@@ -0,0 +1,43 @@
...
+    (deref [_] (f))))
+
+(defn add-on-dispose!
+  "No-op in JVM Clojure, since for testing purposes, we don't care about
+  releasing resources for efficiency purposes."
+  [a-ratom f]
```
danielcompton: Would it be good to be able to dispose? Not sure as they won't be disposed by views?

src/re_frame/loggers.cljc
```diff
@@ -0,0 +1,36 @@
...
+
+;; "loggers" holds the current set of logging functions.
+;; By default, re-frame uses the functions provided by js/console.
+;; Use set-loggers! to change these defaults.
+
+(def ^:private loggers
+  (atom {:log       #?(:cljs (js/console.log.bind   js/console)
+                       :clj  println)
+         :warn      #?(:cljs (js/console.warn.bind  js/console)
+                       :clj  (partial println "WARN: "))
```
danielcompton: Can we use `clojure.tools.logging` here? That'll integrate with the rest of the logging world.

src/re_frame/router.cljc
```diff
@@ -224,11 +224,12 @@
...
         ;; See https://github.com/Day8/re-frame/issues/164#issuecomment-233528154
-        stack  (->> (js/Error. (str "Event " (first event-v) " dispatched from here:"))
-                    .-stack
-                    clojure.string/split-lines
-                    (remove #(re-find #"react.inc.js|\(native\)" %))
-                    (clojure.string/join "\n"))]
+        stack  #?(:cljs (->> (js/Error. (str "Event " (first event-v) " dispatched from here:"))
+                             .-stack
+                             clojure.string/split-lines
+                             (remove #(re-find #"react.inc.js|\(native\)" %))
```
danielcompton: This wasn't you, but probably good to fix the regex to `#"react\.inc\.js|\(native\)"`

src/re_frame/interop.clj
samroberton: I don't like the idea of calling `f` directly because it would mean that in your tests you would see synchronous behaviour that in the real-world would be synchronous.  I think that makes tests a bit less likely to find bugs.  Probably the best thing would be just to use `future` -- the only reason I didn't was I couldn't be bothered double-checking at the time whether it's semantics for exceptions thrown by `f` are sensible for this use-case.  Let me double-check that.

src/re_frame/interop.clj
samroberton: Personally, I think it's better not to try do that at all.

Fundamentally, unless we actually re-implement the whole Reagent and React component tree rendering lifecycle (invoking Reagent render functions only when Reagent would, etc), I think it's going to be almost impossible to ensure we `deref` ratoms in tests only in the same way as Reagent would in the real world.  If we implement some form of `IReactiveAtom` and `IDisposable` regardless, then I think that's more likely to mislead people into thinking that we *are* accurately mimicking real-world behaviour, and writing bogus tests.

I think the naive implementation is fine, though, because as described in the docstring here, in most cases that won't make a difference.  For cases where it *would* make a difference, I think we just need to say "you can't test that type of behaviour with this test code".

Having said that, the sort of code I'm thinking of testing with this is all code for which this doesn't make a difference.  If you have some examples of code which would be not be accurately testable with the above, but which would become accurately testable with some additional cleverness in this `ns`, I'd be interested to work through what might be achievable here.

Regardless, at the very least I think your comment is enough to suggest that an more complete explanation for the approach this namespace takes is probably warranted.

src/re_frame/interop.clj
samroberton: Are there particular examples you can think of where the correctness of a test would depend on whether something was disposed or not?  Or cases where failing to dispose of something would be problematic for other reasons?

src/re_frame/loggers.cljc
samroberton: We could.  I didn't on the assumption that in tests you probably don't care, and I didn't want to introduce a new dependency -- even if it is as harmless as `tools.logging`.  (Especially given that re-frame has obviously chosen not to use a ClojureScript logging framework even in prod code, presumably for a similar reason.)  Is there a benefit to using tools logging in tests rather than just `println`?

src/re_frame/interop.clj
samroberton: I'm unsure what the best way to approach this is.  Creating a new protocol and extending it to `IAtom` would mean that we don't detect when you've just passed in a raw `clojure.core/atom` -- in which case there's no point in a protocol, we might as well just test whether you've passed in an atom.

What I've done here seems like the easiest way to detect whether you've used `re-frame.interop/ratom` to create the atom.  But I'm not sure how useful that check actually is in the first place.  The `ratom?` function gets used to sanity-check that `pure`, `subscribe` and `fx` get given `app-db`, not some other random thing.  I vaguely wonder whether in testing, it's more appropriate just to say "yup, I got some sort of atom, that's good enough" and leave the Reactive-ness part of the check to the CLJS runtime?

So I'm inclined to replace the CLJ implementation of `ratom` with just `(atom x)`, getting rid of the whole metadata thing.

src/re_frame/interop.clj
samroberton: Actually, `future` is not the right answer (but nor is `(.start (Thread. f))`.

Given that the purpose of this function is for `re-frame.router/event-queue` to manage the dispatch of events, and given that CLJS is single-threaded, it would be kind of silly to have the tests multi-threading the processing of dispatched events.  I'll change this to use a single-thread executor.

src/re_frame/interop.clj
samroberton: That said, I could see that implementing `IRef` (probably by extending `clojure.lang.ARef`) to get watch support might be necessary for some uses.

I think that probably makes sense as a future enhancement, though, rather than trying to guess every thing that might be required right now.

src/re_frame/loggers.cljc
danielcompton: I think it's probably better to use tools.logging so that you can control logging levels. It could be used either for certain tests that you know will cause an error to be logged that you don't want to be emitted, or just controlling the overall logging level at different times. Otherwise, you will get println's in your tests that you can't turn off.

danielcompton:
This looks pretty good to me. I think it probably makes sense to bring in the minimal implementation at this point, rather than trying to replicate Reagent entirely on the JVM, and as more features are needed, they can be added.

Probably the only other thing that would be good to have is some simple tests in Clojure that exercise this all, as well as providing guidance for others on how to run tests in CLJ. You can also add `lein test` to the `circle.yml` so that they get tested in CI, which should help prevent regressions.

src/re_frame/interop.clj
danielcompton: I'm mainly thinking of testing the dispose method itself in the test, i.e. ensuring that all resources are cleaned up when dispose is called. But that doesn't need to be added now.

project.clj
```diff
@@ -7,12 +7,13 @@
...
   :profiles {:debug {:debug true}
-             :dev   {:dependencies [[karma-reporter     "0.3.0"]
-                                    [binaryage/devtools "0.7.2"]]
-                     :plugins      [[lein-cljsbuild     "1.1.3"]
-                                    [lein-npm           "0.6.2"]
-                                    [lein-figwheel      "0.5.4-7"]
-                                    [lein-shell         "0.5.0"]]}}
+             :dev   {:dependencies [[karma-reporter            "0.3.0"]
+                                    [binaryage/devtools        "0.7.2"]
+                                    [org.clojure/tools.logging "0.3.1"]]
```
stumitchell: I moved this out of dev dependencies, as needs to be pulled in in consuming projects

PR #186, opened 2016-07-22
Fix never-ending warn about not implementing IReactiveAtom

oliyh:
Currently whenever you have the console open you get a warning about some arguments not implementing IReactiveAtom. After a lot of time trying to track down where the problem lay I realised that my code was fine, and the warning is always printed with an empty list.

Hope this helps
Oliy

### DIFF
--- a/src/re_frame/subs.cljs
+++ b/src/re_frame/subs.cljs
@@ -79,7 +79,7 @@
      (let [query-id   (first-in-vector v)
            handler-fn (get @qid->fn query-id)]
        (when ^boolean js/goog.DEBUG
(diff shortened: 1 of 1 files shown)


mike-thompson-day8:
Many Thanks!

src/re_frame/subs.cljs
```diff
@@ -79,7 +79,7 @@
      (let [query-id   (first-in-vector v)
            handler-fn (get @qid->fn query-id)]
        (when ^boolean js/goog.DEBUG
-         (when-let [not-reactive (remove #(implements? reagent.ratom/IReactiveAtom %) dynv)]
```
danielcompton: I did wonder about this when I read the code, but assumed that when-let did nil punning.

PR #187, opened 2016-07-26
Update logo

martinklepsch:
- No more padding around logo
- "f" is now more balanced/symmetric
- additional sizes
- use more rounded ubuntu font


mike-thompson-day8:
Thanks again Martin!!

PR #189, opened 2016-08-08
Bugfix: path and fx handler

samroberton:
If there is no `:db` effect produced by an effectful handler (presumably
because the handler is concerned with some other form of side-effect
than a DB update), then the `path` interceptor should not attempt to
update the DB value it returns.

### DIFF
--- a/src/re_frame/std_interceptors.cljc
+++ b/src/re_frame/std_interceptors.cljc
@@ -164,11 +164,11 @@
                  (let [db-store     (db-store-key context)
                        original-db  (peek db-store)
                        new-db-store (pop db-store)
-                       full-db      (->> (get-effect context :db)
-                                         (assoc-in original-db path))]
-                   (-> context
-                       (assoc db-store-key new-db-store)
-                       (assoc-effect :db full-db)))))))
+                       new-context  (assoc context db-store-key new-db-store)]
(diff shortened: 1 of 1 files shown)


mike-thompson-day8:
@samroberton a test was failing (false looked like a nil) so I created this alternative fix https://github.com/Day8/re-frame/commit/6a172ddbe06788591affdf84fd90ab0f8ac081ee
Does it cover the problem usecases you identified?

samroberton:
@mike-thompson-day8 your commit is definitely a better fix (and yes, it does fix the issue I observed in our codebase) -- especially in that it also allows you to return `nil` to indicate that you want this whole path to be set to `nil`.

One suggestion, though:  with this facility now available, the `debug` interceptor should use the new arity as well (on `std_interceptors.clj` lines 38 & 39).  One of the reasons this bug took me a while to track down was that the `debug` interceptor was telling me that there was no DB change happening, then suddenly bits of my DB were ending up `nil` -- ie inconsistency between what `debug` reports is happening, and what actually happens.

samroberton:
Closing since 6a172dd covers the original bug, and 486c7e6 covers the suggestion to get the `debug` interceptor to behave the same way.

Thanks for this, Mike.

PR #200, opened 2016-08-22
Remove trailing spaces from console logging

danielcompton:
console.log and family provide spaces between arguments, so we don't need to add them in our console errors. This is how it looks currently: 

screenshot of google chrome 22-08-16 4-10-14 pm


mike-thompson-day8:
Thanks

PR #202, opened 2016-08-23
Missing localstore query

tomasd:
cofx should be queried by :localstore and then :defaults-key


mike-thompson-day8:
Thanks for the feedback!   It looks as though this issue has already been edited away in subsequent work.  Closing.

PR #206, opened 2016-08-23
Update "A Larger App" documentation and docs/readme

smahood:
This PR might need some tweaking, I've changed a few things in the structure that may not meet the goals of the docs.

Content changes 
- added in "reg-event-db" and "reg-sub" to code section
- added re-frame namespace to function calls (eg. "dispatch" to "re-frame/dispatch") to make it clear where those functions are defined
- tested code with re-frame 0.8.0 and fixed a typo (missing `]` in `let`)

Other (perhaps not ideal) changes 
- Added "Structuring Your Application" section to docs/README.md
- Renamed "A-Larger-App.md" to "Basic-App-Structure.md"
- Split out the panel navigation section into "Navigation.md"
- Split out the synthetic keyword section into "Namespaced-Keywords.md"
- Added links to allow readers to follow through the 3 documents in order

Reasoning for the changes 
- The existing docs cover both smaller and larger apps and the current title doesn't make that clear
- Seems like many people (me included) read the original form of the documentation and missed the section on synthetic keywords
- I think the Navigation document would be an ideal place to add info or links about URL routing such as Secretary or to add links to existing info

Rather than splitting the docs up, I think it would be pretty reasonable (and closer to the current docs) to combine them into one document with a TOC and name it "Structuring Your Application" or something similar - if that's preferable I can edit and submit another PR.


mike-thompson-day8:
Love it.  Thanks!

PR #211, opened 2016-08-25
First try at updating "Subscribing to a Database" docs

smahood:
Mainly just a syntax update, changing `register-sub' and adding a bit of explanation to a couple areas.

I think it needs more work, but this will at least give us a base to discuss. 

Questions: 
- Can you check out the explanation for why we are using `reg-sub-raw` and see if it's accurate? Should we be using 'reg-sub-raw' there? 
- What needs to be done to update the undo-redo information? Is there a new link?
- Is there anything we should add in terms of effectful handlers or that is specific to the http-fx library as in https://github.com/Day8/re-frame/blob/master/docs/Talking-To-Servers.md ?


mike-thompson-day8:
@smahood I merge this and do a bit of massage.  Thanks!

PR #212, opened 2016-08-26
A bit of clean up work

smahood:
Fixed some typos.
Made spacing more consistent in `require:` sections.
Minor wording change in router.cljc in queue explanation.
Changed begatted to begotten in router.cljc.


mike-thompson-day8:
So many typos.  Thanks!!

PR #215, opened 2016-08-28
Correct spelling

mbertheau:
https://en.wikipedia.org/wiki/Homage_(arts)


mike-thompson-day8:
Thanks.  I included that change manually in the most recent iteration (will they ever end).

PR #221, opened 2016-08-30
Add missing handler to "Loading Initial Data" doc

si14:
"The pattern" implies `:initialise-db` handler, but it is missing. This PR adds a very simple implementation.


mike-thompson-day8:
Thanks for those three fixups !!

PR #227, opened 2016-09-05
Add elfeed-cljsrn app to docs/External-Resources

areina:
Hi!

I would like to add this application to the external-resources document. Maybe it could be useful as a reference for people learning re-frame, or trying to build an application with React Native and Clojurescript. 
It's based on re-frame 0.8.0 and as interesting points, it's using `async-flow-fx` and `http-fx` effect handlers.

Thank you!

### DIFF
--- a/docs/External-Resources.md
+++ b/docs/External-Resources.md
@@ -18,6 +18,8 @@ Templates, examples, and other resources related to re-frame. For additions or m
 
 ### Examples and Applications Using re-frame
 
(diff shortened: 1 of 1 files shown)


smahood:
Hi @areina, I've added these changes manually to the develop branch (I'm not sure how to change where your PR points to so this way is makes it less likely for me to break something). 

Thanks!

danielcompton:
For the future, there's a new GitHub feature which lets you change the branch a PR points to:

screenshot of safari technology preview 5-09-16 4-29-12 pm

PR #235, opened 2016-09-12
Fix typo in first effect map

micmarsh:
Considering it results in an invalid map literal, I'm assuming that `1` wasn't intended


docs/Solve-the-CPU-hog-problem.md
```diff
@@ -57,7 +57,7 @@ Here's an `-fx` handler which counts up to some number in chunks:
       ;; We are at the beginning, so:
       ;;     - modify db, causing popup of Modal saying "Working ..."
       ;;     - begin iterative dispatch. Give initial version of "so-far"
-      {:disptch 1[:count-to false {:counter 0} finish-at]  ;; dispatch to self
+      {:disptch [:count-to false {:counter 0} finish-at]  ;; dispatch to self
```
mbertheau: Should this also be `dispAtch`?

micmarsh: Good catch, fixed

smahood:
Thanks!

Issue #232, opened 2016-09-08
Broken links in docs


deg:
At the end of [https://github.com/Day8/re-frame/wiki/Debugging](https://github.com/Day8/re-frame/wiki/Debugging), there is a section "Macro Help".  The links in it are all broken, apparently because refheap.com has gone non-responsive.

mike-thompson-day8:
Quick Note:  all Wiki documentation is being transferred to within the `/docs` folder.  
https://github.com/Day8/re-frame/blob/master/docs/Debugging.md

But, yes,  those links are currently down due to a server 500 error.  I'm assuming they will return soon (they worked a week ago).

deg:
Rumor has it (as heard on the #clojure IRC chanel) that the refheap site is down permanently.

I just tried searching on archive.org too, but no luck.

mike-thompson-day8:
@mccraigmccraig your refheap links [at the bottom of this document](https://github.com/Day8/re-frame/blob/master/docs/Debugging.md)  have gone south.   It looks like refheap has gone dark. 

Any chance you can supply some alternative ones?  Gists perhaps?

si14:
I've just added a PR to inline those snippets #238

mccraigmccraig:
ha, awesome @si14 - i wasn't doing very well at paying any attention to this !

smahood:
PR has been merged to `Develop` branch.

PR #238, opened 2016-09-27
Inline refheap snippets

si14:
Refheap links are unreliable, so it's better to inline the linked code (see #232)


smahood:
Thanks!

PR #247, opened 2016-10-05
Add project readme to docs table of contents

smahood:
@mike-thompson-day8 - Not sure this is the best way to do it, but I think it's worth having a link to the main project readme in the docs.


danielcompton:
Thanks!

PR #249, opened 2016-10-13
Add :dispatch-debounce fx handler

### LINKED ISSUE #233: How to debounce or delay a dispatch?
Sorry for the ignorance ... what's the best way to `debounce` or `delay` a dispatch? In redux land there are various kinds of middleware for [that](https://github.com/ryanseddon/redux-debounced).

Maybe you could add an answer to this to the FAQ section?!

danielcompton:
Some code from @rafd used (thanks!).

Fixes #233

### DIFF
--- a/docs/Effects.md
+++ b/docs/Effects.md
@@ -26,6 +26,7 @@ make side effects a noop in event replays.
     + [:dispatch-later](#dispatch-later)
     + [:dispatch](#dispatch)
     + [:dispatch-n](#dispatch-n)
+    + [:dispatch-debounce](#-dispatch-debounce)
     + [:deregister-event-handler](#deregister-event-handler)
     + [:db](#db)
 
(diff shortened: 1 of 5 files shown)


danielcompton:
@rafd, I used some of your code, as a base here I hope you don't mind?

rafd:
@danielcompton No problem! :) Thanks for doing all the work on this.

rafd:
Since you are using core.async, you can use it to implement debounce instead of  `js/setTimeout` (with the added bonus of not having to use the cljs reader conditionals).

There are a number of `debounce` core.async implementations floating around:

 - https://github.com/swannodette/async-tests/blob/master/src/async_test/utils/helpers.cljs#L196
 - https://github.com/swannodette/swannodette.github.com/blob/master/code/blog/src/blog/utils/reactive.cljs#L279
- https://gist.github.com/xfsnowind/e15cc2e6da74df81f129
- https://github.com/Day8/re-com/blob/master/src/re_com/typeahead.cljs#L301
- https://github.com/metosin/metosin-common/blob/672f547c76ab381a85f8e1ac92e30b6ac08c8ca9/src/cljc/metosin/core/async/debounce.cljc

Here is the one extracted from one of my projects:

```
(defn debounce
  "Given the input channel source and a debouncing time of msecs, return a new
  channel that will forward the latest event from source at most every msecs
  milliseconds"
  [source msecs]
  (let [out (chan)]
    (go
      (loop [state ::init
             lastv nil
             chans [source]]
        (let [[_ threshold] chans]
          (let [[v sc] (alts! chans)]
            (condp = sc
              source (recur ::debouncing v
                            (case state
                              ::init (conj chans (timeout msecs))
                              ::debouncing (conj (pop chans) (timeout msecs))))
              threshold (do (when lastv
                              (put! out lastv))
                            (recur ::init nil (pop chans))))))))
    out))
```

rafd:
Ah, I just noticed you are only using core.async for testing. You can ignore my previous comment, as it may not be worth adding a core.async dependency for the actual project.

docs/Effects.md
```diff
@@ -323,6 +324,37 @@ usage:
...
+Repeated dispatches will reset the timer. Use this in scenarios where you want to wait for quiescence
+before dispatching.
+
+Expects either a single map, or a sequence of maps.
+
+usage:
+
+```cljs
+{:dispatch-debounce [{:id       ::re-render-markdown
+                      :wait     250
```
jamesnvc: This should be `:timeout`, not `:wait`

danielcompton:
This needs to go into a library. If anyone would like to make a library out of this, I'm happy to donate the code. Possibly it would go along with @kp2222's changes in #258?

avocade:
+1 for a library ☺️

thegeronimo:
Has this ever made it into a library?

PR #251, opened 2016-10-13
Added :response-format, without that it fails.

kapware:
Without this line the event fails as the response is not properly formatted.


smahood:
Good catch, thanks!

PR #252, opened 2016-10-15
test logging of missing effect handlers

kp2222:


danielcompton:
Hey thanks for this, it was pretty close. I was wanting to check for missing fx handlers, not missing event handlers, but the test was basically the same. Thanks!

PR #254, opened 2016-10-17
Add dynamic var for warning on handler overwrite

### LINKED ISSUE #204: Don't warn about reloading handlers when initiated by figwheel
There may be a way to tell re-frame when figwheel is reloading code so that we can ignore the warnings for overwriting handlers. This would reduce the noise in the console, and make it more likely we notice actual cases of this happening.

danielcompton:
When figwheel reloads code, it causes handlers to be re-registered which leads to a warning for each handler. This leads to noisy logs.  Figwheel has a :before-jsload key which can be called before it reloads application code.

This patch adds a `*warn-on-overwrite*` dynamic var, and shows how to use it in the example project.

One side effect of using this is that you will never get a warning for duplicate handlers when Figwheel is reloading. It may be possible to track how many handlers were reloaded in a particular Figwheel reload, but this would get very complex. In any case, you will still get warnings every time you refresh the browser, which should be good enough for the relatively rare case of creating duplicate handlers.

This still needs 
- [ ] tests
- [ ] a decision on where to put the dynamic var (in `re-frame.core` or elsewhere?, if in core, what about circular dependencies?)
- [ ] Recommendations on the [usage of dynamic vars](https://groups.google.com/forum/#!topic/clojurescript/0FdQB4bwtSw)
- [x] Is `:before-jsload` supported/able to be used? 
- [ ] Update changelog
- [ ] Docs on usage

Fixes #204

### DIFF
--- a/examples/todomvc/project.clj
+++ b/examples/todomvc/project.clj
@@ -17,7 +17,8 @@
                                                   :source-map           true
                                                   :source-map-timestamp true
                                                   :main                 "todomvc.core"}
-                                       :figwheel {:on-jsload "todomvc.core/main"}}}}}
+                                       :figwheel {:before-jsload "todomvc.core/before_reload"
+                                                  :on-jsload "todomvc.core/figwheel_reload"}}}}}
 
              :prod {:cljsbuild
                     {:builds {:client {:compiler {:optimizations :advanced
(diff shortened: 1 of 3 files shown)


kommen:
Is there some intention to implement this later?

kennyjwilli:
@kommen It was implemented in this commit https://github.com/Day8/re-frame/pull/254/commits/8de4175c6e8337b457c421b5064a3d86ddf8e10d.

danielcompton:
I actually didn't mean to drop this, I accidentally closed it when I moved stuff over to the master branch.

Issue #216, opened 2016-08-28
Create API documentation for `core.cljs`


mike-thompson-day8:
The API for re-frame has grown to the point that there should be docs. That means docs for the functions exposed in `re-frame.core` 
- [x]  decide which tool to use - codox ?
- [x]  where to put the docs? 
- [x]  review all docstrings for clarity and completeness. 
- [x]  historically, `core.cljs` used `def` to promote functions deeper in the implementation up to the API level.  These defs don't have docstrings.  What to do about this?

That final point describes this practice within `core.cljs`:

```
(def some-thing  re-frame.impl/some-thing)
```

danielcompton:
Relates to #62.

mike-thompson-day8:
See also https://github.com/Day8/re-frame/pull/257

agzam:
As a newbie to re-frame this baffles me a bit. Why there's no unified api docs wiki or whatever? Or everything is really so simple - after you learn the basics, you won't need documentation to any of `re-frame.core` functions?

mike-thompson-day8:
Revising this issue after a long absence.

A good-enough version of API docs can be found here:  https://github.com/Day8/re-frame/blob/master/docs/API.md

I won't close because eventually we'd like a better solution.

superstructor:
Any solution must deliver the following two outcomes: 
- API documentation generation (e.g. codox, cljdoc etc)-
- Common IDEs code assist features (e.g IntelliJ/Cursive, Emacs, Calva etc) 

I investigated 3 options
1. using a macro - like “defvar” proposed in PR #257
2. move the relevant functions in their entirety to re-frame.core
3. create wrapper functions with the docstring and args that delegate to the other namespaces thanks to suggestion by @mike-thompson-day8 

Tools such as IntelliJ/Cursive and Joker perform static analysis of code which rules out any macro-based approach. 

Moving the entire function implementation is undesirable from a code organisation and style perspective. 

Therefore we decided to go with creating wrapper functions with the docstrings and args in re-frame.core.

PR #257, opened 2016-10-19
Add docstrings to vars in re-frame.core

grammati:
Cider is able to show docstrings of clojurescript functions (using `C-d C-d d`), but for most functions in re-frame.core, it just shows "Not documented", which is sad and kind of ironic given that re-frame is probably the best-documented project in all of clojuredom.

This commit adds a tiny macro (that I called `defvar` only because I couldn't think of anything better) which copies over the `:doc` metadata from the implementation function.

### DIFF
--- a/src/re_frame/core.cljc
+++ b/src/re_frame/core.cljc
@@ -1,5 +1,7 @@
 (ns re-frame.core
+  #?(:cljs (:require-macros [re-frame.macros :refer [defvar]]))
   (:require
+    #?(:clj [re-frame.macros :refer [defvar]])
     [re-frame.events           :as events]
     [re-frame.subs             :as subs]
     [re-frame.interop          :as interop]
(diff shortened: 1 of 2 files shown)


danielcompton:
This looks neat! With Cider, what do you need running for this to work? Does it need a CLJS REPL or can it use a CLJ one because re-frame uses CLJC?

grammati:
I tried in both a clj and a cljs repl and was able to get the docs in both.

danielcompton:
Sorry for the delay on this. We're still considering this internally as it is a bit of a change adding a custom macro here.

danielcompton:
Hey, I've delved into this a bit deeper. I had a go at this in a CLJS figwheel REPL, and I couldn't get docs for the defvar'd vars. How did you get the docs in a CLJS repl?

If we're going to make this change, it would also be really handy to get the other metadata from the functions, particularly the arglist so that we can generate docs with codox. I've had a look at getting the other metadata but ran out of time. I suspect that copying how `cljs.repl/doc` works would be the way to do it.

If you want to get docs from Clojure REPL's as well, then you may want to make the macro one that can run from Clojure and ClojureScript via https://groups.google.com/forum/#!topic/clojurescript/HsWTuhMP7yc.

I understand if you don't have the time/inclination to go this deep into it. Sorry for the delay on this side.

superstructor:
We investigated this PR as a solution for #216 and #456.  In the end, we had to reject this approach and resolve the issue another way because tooling like IntelliJ/Cursive or Jocker do static analysis of code and a macro expansion approach won’t work for them.
 
But, many thanks for the PR all the same @grammati! We looked at it carefully and your suggestion certainly helped us to research for this fix.

Closing.

Issue #250, opened 2016-10-13
Add a :dispatch-throttle effect


danielcompton:
Throttle is a sibling to debounce, and they share many similarities. I would gladly take a PR for a `:dispatch-throttle` effect. Requirements:
- Behaves like a throttle, e.g. https://css-tricks.com/debouncing-throttling-explained-examples/, https://lodash.com/docs/4.16.4#throttle
- Takes a map of effects, or a sequence of maps (the same as #249)
- Takes the same map options as `:dispatch-debounce`, with the addition of `:leading?` and `:trailing?` which behave the same as in lodash's [throttle](https://lodash.com/docs/4.16.4#throttle). `timeout` is renamed to `window-duration`
- Tests (the ones written for `:dispatch-debounce` in `re-frame.fx-test` are probably a good start)
- Docs

I'm happy to help with docs and tests, the tricky part which I don't have time for right now is the throttling implementation.

kp2222:
I have created a pull request #258 

The changes are mostly based on #249. Please feel free to suggest anything that needs change

danielcompton:
Closing as we've decided that this probably belongs in a separate lib, more details in https://github.com/Day8/re-frame/pull/258#issuecomment-262658598.

PR #258, opened 2016-10-20
Add a :dispatch-throttle effect #250

kp2222:
These changes are mostly based on a similar pull request for debounced events #249

### DIFF
--- a/docs/Effects.md
+++ b/docs/Effects.md
@@ -346,6 +346,39 @@ usage:
 {:db  {:key1 value1 key2 value2}}
 ```
 
+#### :dispatch-throttle
+
+This will only dispatch events at most once per every window-duration milliseconds.
+The throttled function can be configured to dispatch on the leading, trailing or both edges of this window.
+
+Note: When both leading and trailing flags are enabled the event will be dispatched on trailing edge only if there are more than one event.
+
+usage:
+
+Expects either a single map, or a sequence of maps.
+
+```clj
+{:dispatch-throttle {:id       ::re-render-markdown
+                     :window-duration  250
+                     :leading? true
+                     :trailing? false
+                     :dispatch [:re-render :main-section]}}
+```
+
+Cancel throttled events
+
+```clj
+{:dispatch-throttle {:id       ::re-render-markdown
+                     :action :cancel}}
+```
+
+Flush latest throttled event
+
+```clj
+{:dispatch-throttle {:id ::re-render-markdown
+                     :action :flush}}
+```
+
 ### External Effects
 
 - https://github.com/Day8/re-frame-http-fx   (GETs and POSTs)
--- a/project.clj
+++ b/project.clj
@@ -9,7 +9,8 @@
 
   :profiles {:debug {:debug true}
              :dev   {:dependencies [[karma-reporter            "1.0.1"]
-                                    [binaryage/devtools        "0.8.1"]]
+                                    [binaryage/devtools        "0.8.1"]
+                                    [org.clojure/core.async    "0.2.391"]]
                      :plugins      [[lein-ancient              "0.6.10"]
                                     [lein-cljsbuild            "1.1.4"]
                                     [lein-npm                  "0.6.2"]
--- a/src/re_frame/fx.cljc
+++ b/src/re_frame/fx.cljc
@@ -3,7 +3,7 @@
     [re-frame.router      :as router]
     [re-frame.db          :refer [app-db]]
     [re-frame.interceptor :refer [->interceptor]]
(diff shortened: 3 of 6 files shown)


danielcompton:
Thanks heaps for this, I especially appreciate the extensive tests :) Will take me a little bit to give it a proper review, but hope to get to it soon. If you have time, would you be able to add some docs to Effects.md for this? Otherwise I can add them. Thanks!

kp2222:
Thanks for the quick reply. Here are the docs :)

lsnape:
My expectations are that `dispatch-throttle` defers any event from being handled until after `window-duration` milliseconds has elapsed. But with this implementation that behaviour only holds true if more than 1 event is queued at one time. Unless the events are queued synchronously it is likely that they will always be dispatched immediately. I think you need to persist a `last-ran` timestamp so that you know when the last event occurred.

Thanks for taking the time to write this PR – Looking forward to using this. 👍

kp2222:
what do you mean by 
> "dispatch-throttle defers any event from being handled until after window-duration milliseconds has elapsed. But with this implementation that behavior only holds true if more than 1 event is queued at one time" ?  

Doesn't this test case cover this scenario?

```
(deftest dispatch-throttle-only-trailing
  (async done
         (go
           (let [seen-events (atom [])
                 throttled-event (fn[event]
                                   {:id ::dispatch-throttle-only-trailing
                                    :window-duration 50
                                    :leading? false
                                    :trailing? true
                                    :dispatch [::type-ahead-search event seen-events]})]
             (re-frame/dispatch [::throttle-test (throttled-event "a")])
             (<! (timeout 10))
             (is (= @seen-events []))
             (<! (timeout 60))
             (is (= @seen-events ["a"]))
             (done)))))
```

The behavior of the first event getting dispatched immediately can be turned off by setting leading? to false. Of course, I might be missing something completely here. I am happy to work on the changes once I figure out the issues are :)

lsnape:
OK, understood! I think I've found the source of my confusion: to my mind, throttling is the process by which events are prevented from dispatching too quickly so an not to overwhelm another part of the system. It doesn't drop events, only defers them from running quicker than the interval specified. 

I see now that your implementation is something different: you're only setting the last seen event in the `throttled-events` atom and using `leading?`, `trailing?` and `window-duration` to determine when that should be run.

Would it not make more sense to accumulate the events, rather than only defer the last seen event? After all, it's still possible to cancel older / newer events in the handler of the deferred event itself.

I'd be interested to hear your / anyone elses thoughts! :)

kp2222:
Ah, I see what you were saying now :) However, I can't think of a way to implement event throttling without dropping the events.

The problems I see in implementing something like this is :

* This would require an unbounded queue (so we don't drop any events). Unbounded queues are almost always a bad idea.

* Assuming we keep an unbounded queue of events, as the queue starts getting bigger the recent events will have to wait too long before dispatched and I can't think of  any actual use case where this would be useful.

In addition to this, we were looking at lodash's  implementation of [throttle](https://lodash.com/docs/4.16.6#throttle) as a reference implementation.

danielcompton:
> to my mind, throttling is the process by which events are prevented from dispatching too quickly so as not to overwhelm another part of the system. It doesn't drop events, only defers them from running quicker than the interval specified.

The intent of this effect handler is to provide something similar to lodash's throttle. I can see why you might want something like what you're describing, but this isn't it :)

lsnape:
OK, thanks for bearing with me. I've now implemented my own fx for what I want - really easy to do in the end :)

One last suggestion: do you want to ensure that no throttled events are handled after a `:cancel` action is received? If so then I think a `dispatch-sync` is needed (though not entirely sure what the drawbacks are to this approach).

Cheers

Latros:
Nice work @kp2222, looking forward to using this

danielcompton:
After talking with @mike-thompson-day8 we've decided we'd prefer to keep this and #249 as separate FX handlers that aren't part of re-frame core. They'll still be advertised like other effects in libs so people can learn about them. This is mainly as people have different preferences for these kinds of FX, to reduce the maintenance burden on re-frame, and that they probably don't belong as part of the [naked robotic core](http://5by5.tv/hypercritical/86) of re-frame. 

@kp2222 if you end up putting this PR into a library, I'm happy to add #249 to it as, as debounce and throttle are quite similar.

Sorry for the delay and run-around after inviting PR's for this feature 🙏.

avocade:
+1 for breaking this out in a separate lib.

thegeronimo:
> Add :dispatch-debounce fx handler #249

Did this ever become a library?

PR #268, opened 2016-11-16
Minor tweaks to README.md

TashaGospel:


danielcompton:
@mike-thompson-day8 will let you make the call on these doc tweaks.

mike-thompson-day8:
@TashaGospel I really appreciate all input and would normally be happy to accept your suggestion  ... except I've just finished a rewrite on the README.  So the words you have tweaked are not there any more.

PR #272, opened 2016-12-01
Typos

Latros:
Just reading through each namespace & trying to wrap my head around things so I can give a proper contribution a shot. Noticing a few typos along the way :)


Latros:
Will re-submit once I finish up

PR #282, opened 2016-12-08
Updated previous to go to correct page and fixed index 404

assoc-in:
I found a couple link problems when I was reading through the docs. Keep up the great work on re-frame!


mike-thompson-day8:
Thanks!

PR #283, opened 2016-12-10
making trim-v before/after symmetrical

oliyh:
Clone of #280 for `develop` branch

Hello,

This change will put the untrimmed event vector back in the after phase, making it behave like the original middleware. This is generally a nice thing for interceptors to do, making them additive and not mutating, or at least idempotent.

This then allows other interceptors in the after phase to work better - middleware interceptors generally want to know what the event is called, because they're generic. For example, it lets me write the following improved schema check interceptor:

(defn valid-schema?
  "validate the given db, writing any problems to console.error"
  [db [event-name]]
  (let [res (s/check schema db)]
    (if (some? res)
      (.error js/console (str "schema problem after event " event-name ":" res)))))
And use it in a stack of interceptors containing trim-v, as long as it's at the end (which it generally will be, because it cleans up the signatures of my event handlers but I don't want it for my middleware interceptors):

(def my-interceptors [(after valid-schema?)
                      trim-v])

### DIFF
--- a/src/re_frame/std_interceptors.cljc
+++ b/src/re_frame/std_interceptors.cljc
@@ -63,7 +63,14 @@
     :id      :trim-v
     :before  (fn trimv-before
                [context]
-               (update-in context [:coeffects :event] subvec 1))))
+               (-> context
+                   (update-in [:coeffects :event] subvec 1)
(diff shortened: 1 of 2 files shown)


danielcompton:
Thanks! Merged in https://github.com/Day8/re-frame/commit/ff7c175de0427eb384667a4e8c345c02646c15ad.

PR #287, opened 2016-12-15
README, add reason to consider adopting re-frame

devinrhode2:


mike-thompson-day8:
Thanks, I like it.

devinrhode2:
This is like one of the fastest merges I've gotten for a simple doc edit!
:)

devinrhode2:
Nice expansion! Didn't know re-frame and elm have such a history

PR #290, opened 2016-12-16
Add re-thread to External Resources doc

milt:
As requested, added [re-thread](https://github.com/yetanalytics/re-thread).


mike-thompson-day8:
Thanks!

PR #293, opened 2016-12-17
fix several typos in docs;

gadfly361:
Note: I added a `TODO` in `docs/Debugging-Event-Handlers.md`, I think there was a clipped sentence at the end.


mike-thompson-day8:
That sort of proofing reading is gold.  Many Thanks.

PR #295, opened 2016-12-18
Missing square brackets on subscription for :name

assoc-in:
I found a missing set of square brackets for a subscription and figured that fixing it will help newer users with the syntax being consistent.


mike-thompson-day8:
Many Thanks. Cumulatively, all these corrections count.

PR #306, opened 2017-01-06
Adding re-learn link to libraries section

oliyh:
There were some weird line endings in the file as well, I cleaned them up!


mike-thompson-day8:
Thanks!

PR #307, opened 2017-01-14
Bikeshed using vector destructuring instead of a let and second.

kenrestivo:
Sorry, it bothered me.

Would totally understand if you don't want it, but couldn't just let it be.


danielcompton:
Thanks for the PR, but we're probably happy with this as it is.

mike-thompson-day8:
Yeah, I wanted to make clear that the 2nd argument was the `event`.  Getting that into a readers head seemed the most useful thing at this point in the teaching process.   Luxuriating in the comforts of destructuring is something for later in the learning curve.

PR #308, opened 2017-01-15
Minor verb changes in README

richardharrington:
Excellent README, very clear! I just have a couple small changes:

1. __In one place, I changed "effected" to "affected"__. The object of the clause is the state, not the change itself, so the verb should be "affected", not "effected".

2. __In another place, I changed "actions" to "puts into action".__ The word "actions" is not a verb. The word "effects" would work here but would make the sentence sound super awkward, so I went with "puts into action".


mike-thompson-day8:
Thanks!

richardharrington:
No problem!

PR #309, opened 2017-01-16
added nested Intellij project files to .gitignore

richardharrington:
For if you want to start an intellij project, for instance, just in the `examples/simple/` folder, which will place a `.idea` folder at that level in the hierarchy.


danielcompton:
Thanks!

PR #313, opened 2017-01-16
Link text to cljs-devtools matchs link

TheTedHogan:
The link text and the link didn't match for cljs-devtools in FAQ 1 : )


mike-thompson-day8:
Thanks!

PR #316, opened 2017-01-19
[docs] Add missing right bracket in the handling example

boechat107:
A simple `]` was missing in the declaration of the anonymous function
to handle `:set-flag` event.


danielcompton:
Thanks, and sorry for the delay.

PR #320, opened 2017-02-01
Update doc table of contents with latest doctoc (1.2.0)

plexus:
It seems the version of doctoc that was previously used didn't correctly handle backticks
and trailing special characters in title names, e.g. it would generate

    [What Are They?](#what-are-they-)
    [`inject-cofx`](#-inject-cofx-)

Instead of what Github expects:

    [What Are They?](#what-are-they)
    [`inject-cofx`](#inject-cofx)


mike-thompson-day8:
Thanks!

PR #325, opened 2017-03-11
self-host compatibility using macrovich

viebel:
This PR makes `re-frame` self-host compatible.
It means that `re-frame` can now be used inside KLIPSE  (which is going super useful for sharing bugs, ideas and more...).
#
Here is a [live demo](http://app.klipse.tech/?container&cljs_in.gist=viebel/e82b197907a7117f72f1a0b9a4a56303).

Note: The self-host compatibility of `re-frame` depends on the self-host compatibility of `re-agent` that is achieved by   https://github.com/reagent-project/reagent/pull/283
image
.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -5,6 +5,7 @@
   :dependencies [[org.clojure/clojure        "1.8.0"]
                  [org.clojure/clojurescript  "1.9.227"]
                  [reagent                    "0.6.0"]
+                 [net.cgrand/macrovich "0.2.0"]
                  [org.clojure/tools.logging  "0.3.1"]]
 
   :profiles {:debug {:debug true}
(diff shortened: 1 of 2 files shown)


arichiardi:
Awesome! Also very good for using experimenting with `lumo`

danielcompton:
Thanks for this, I'll need to review this carefully, but it should be ok. Here's the changes without [whitespace diffs](https://github.com/Day8/re-frame/pull/325/files?w=1).

viebel:
Any update on this PR @danielcompton ?

danielcompton:
I'm ok to merge this for now, on the proviso that if it causes any problem in future versions of Clojure/ClojureScript then it'll be pulled out. So you can use it for things, and I don't anticipate us removing it, but at the moment supporting self-hosting is a non-goal, so don't treat it as being officially supported by re-frame.

danielcompton:
Thanks for this!

PR #327, opened 2017-03-25
Fix for example: Request format is mandatory

schnipseljagd:
The example in Talking-To-Servers.md is missing an mandatory parameter.

Like documented in the example: https://github.com/Day8/re-frame-http-fx#step-2-registration-and-use


danielcompton:
Thanks!

PR #328, opened 2017-03-25
Fix update-coeffect

nberger:
Similar to assoc-coeffect, update-coeffect has to do an update-in `[:coeffects key]`


danielcompton:
Merged in https://github.com/Day8/re-frame/commit/fa7828375ac3149f108662c98e34b91416fd6127. Thanks!

Issue #231, opened 2016-09-07
Detecting Exceptions In Interceptors


mike-thompson-day8:
> The following was raised via Google groups and transferred to here for discussion ... 

Hello, I recently updated to Re-frame version 0.8.0. I have transitioned the majority of my custom middleware to interceptors but am having trouble with one.  I created an application specific logging utility that sends information to Elasticsearch. The error logging middleware looked as follows

``` clj
(defn log-error
  [handler]
  (fn log-error-handler
    [db v]
    (try
        (handler db v)
        (catch :default e
          (do
            (error (.-message e) (.-stack e))
            (throw e))))))
```

This middleware was being included with a few others:
(def standard-middlewares [log-error log-params trim-v])

But now that I am not calling the handler function directly I am unsure how to catch and log errors from an interceptor. Could anyone please point me in the right direction?

mike-thompson-day8:
At the most basic level, you can simply hook `windows.oerror`  to know when exceptions happen during the execution of an interceptor chain.

If we were to adopt the approach used in Pedistal, it would look like this:
   - re-frame catches any exception thrown during the execution of an interceptor chain, including the handler.
   - the exception is assoc-ed into `context`  with the key `:ex`
   - Interceptors can have a `:before`, `:after` and `:error` fn.  
   -  when an exception occurs,  the further execution of interceptors in the chain is stopped, and the interceptor stack is unwound in reverse order, but instead of `:after` being called on each interceptor `:error` is called with the thrown exception.  
   - `:error` fns take one exception argument, and return a new, replacement exception.  They have the option of returning the given argument. 

I've avoided going all-in on exceptions because it does make the process more complicated.  I wanted to better understand the usecases.  If you have one, can you add it to this ticket please.

mike-thompson-day8:
I'm wondering if this approach might be good enough (untested)
```clj
(def log-error
    (let [old-onerror-fn  (atom (nil) 
           new-on-error (fn [e]  
                                      (error e)  
                                     (@old-onerror-fn e)]
       (->interceptor 
           :before   (fn  [context]
                            (reset! old-onerror js/windows.onerror)
                            (oset js/windows.onerror   XXXX)
                            context)
           :after   (fn [context]
                           
    
```

lwhorton:
[...]

PR #330, opened 2017-03-27
Add interceptor error handling

nberger:
As discussed in #231, this allows for handling exceptions as part of the interceptor chain. It's inspired (the same as the overall interceptors implementation) by how pedestal handles exceptions in its interceptors. It gives the possibility to handle exceptions as early as possible (as opposed to a window.onerror handler), with much as context as possible


danielcompton:
Thanks for this. I've taken a first pass over this and it looks pretty good. I'm still digesting the tests, and noodling on introducing the `:error` stage. I also need to take another look at the tests to make sure I grasp everything that's going on there. Have you thought about how this might be described in the docs? I'm not asking you to redo it all now, as we may still go another way on this, but I'm curious as to how we document it.

The main use case I've heard for this so far is for error reporting. re-frame doesn't really encourage using exceptions for flow control in quite the same way as Pedestal's interceptors do, nor do we really intend to handle exceptions from event handlers and continue executing an event. I'm wondering about a more limited approach here instead:

1. Register an error reporter, e.g. `re-frame.core/register-event-error-handler`
2. Run all interceptors in a try/catch
3. If an error occurs, call the error reporter with an exception created by your `exception->ex-info` and abort event processing

I'm aware this approach is a little more primitive, but it doesn't change the interceptor pipeline by introducing a new stage concept. Would this work for your use case? What's your thought on this alternate proposal?

nberger:
@danielcompton: thanks for taking a look.

I actually like your alternative "limited" approach. It would work well for us. And I agree, error reporting/logging seems to be the common use case. The semantics of this error handler would be much simpler so it would also be easier to describe in the documentation. Even though I'll keep thinking about it, I don't have any objection to following that path.

nberger:
@danielcompton I'm still interested on this, but couldn't find the time to work on it lately since I haven't been using re-frame that much. I could try and resolve the conflicts here but the idea is to implement the more simple approach that you proposed, right?

I still think this is an important feature: having a clean way to catch exceptions in the interceptors/handler chain. Even though I haven't been following the latest additions I think this is somewhat missing, right?

nberger:
Closed in favor of #423

Issue #326, opened 2017-03-16
Enforce reg-sub :<- syntax


danielcompton:
Currently a `reg-sub` can have any keyword instead of `:<-` to indicate input signals. For example:

```cljs
(re-frame/reg-sub
  :app/second-sub
  :-> [:app/my-cool-sub]
  (fn
    [cool _]
    cool))
```

is perfectly valid. We should enforce `:<-` so that we have the option to use the syntax in the future.

mike-thompson-day8:
@danielcompton  I believe we can close this?

danielcompton:
Yep

PR #336, opened 2017-04-14
enforce reg-sub :<- syntax

Sohalt:
This implements #326

### DIFF
--- a/src/re_frame/subs.cljc
+++ b/src/re_frame/subs.cljc
@@ -183,18 +183,18 @@
                              f)
 
                          ;; one sugar pair
-                         2 (let [[label vec] input-args]
-                             (when-not (= :<- label)
-                               (console :error err-header "expected :<-, got:" label))
+                         2 (let [[marker vec] input-args]
+                             (when-not (= :<- marker)
+                               (console :error err-header "expected :<-, got:" marker))
                              (fn inp-fn
                                ([_] (subscribe vec))
                                ([_ _] (subscribe vec))))
 
                          ;; multiple sugar pairs
-                         (let [pairs (partition 2 input-args)
-                               labels (map first pairs)
-                               vecs  (map last pairs)]
-                           (when-not (and (every? #{:<-} labels) (every? vector? vecs))
+                         (let [pairs   (partition 2 input-args)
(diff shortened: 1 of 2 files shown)


src/re_frame/subs.cljc
```diff
@@ -183,14 +183,18 @@
                              f)
 
                          ;; one sugar pair
-                         2 (fn inp-fn
-                             ([_] (subscribe (second input-args)))
-                             ([_ _] (subscribe (second input-args))))
+                         2 (let [[label vec] input-args]
+                             (when-not (= :<- )
```
mike-thompson-day8: Shouldn't this be:   `(when-not (= :-< label))`

Sohalt:
Yes, obviously. I don't know how I did miss that. Thanks for catching it.

Sohalt:
I amended the fix. Does the rest look ok to you?

danielcompton:
This looks pretty good. I've got some style tweaks and I might add some tests for this behaviour, but it looks sound. Thanks!

src/re_frame/subs.cljc
```diff
@@ -183,14 +183,18 @@
                              f)
 
                          ;; one sugar pair
-                         2 (fn inp-fn
-                             ([_] (subscribe (second input-args)))
-                             ([_ _] (subscribe (second input-args))))
+                         2 (let [[label vec] input-args]
```
danielcompton: `vec` should be `query-v` to not shadow `cljs.core/vec`

danielcompton: I'm not that keen on `label` here, but I'm struggling to think of a different name. Perhaps `directive`?

mike-thompson-day8: beacon? marker? glue? sentinel?

mike-thompson-day8: delimiter?

mainej: sigil? token?

Sohalt: Any consensus? Out of the suggestions I personally like `marker` the most.

danielcompton:
Thanks @Sohalt, I've merged it with `marker` for now.

PR #338, opened 2017-05-12
Removed duplicated require causing build error with shadow-build

ggeoffrey:
This double `:require` was causing an error and blocking the build when compiling cljs with [Shadow Build](https://github.com/thheller/shadow-build)
> ExceptionInfo NS:re-frame.interceptor has duplicate require/use for re-frame.interop  clojure.core/ex-info (core.clj:4725)  

Merging them into one fixed it.


danielcompton:
Thanks!

danielcompton:
I've just released `[re-frame 0.9.3]` which includes this fix.

PR #341, opened 2017-05-18
Don't run on-changes if db is not in effects

bradleesand:
Currently, if there the effect handler does not set `:db`, then `on-changes` will end up treating `new-db` as `nil` and probably end up wiping the `db`. This causes the `on-changes` interceptor not to run if the `:db` is not set in the effect.


mike-thompson-day8:
Many Thanks, good fix.

PR #347, opened 2017-05-31
fix requires

cmal:
I found `re-frame` use some clojure official libraries with the unstrict format, that make it impossible to be used with some tools like `shadow-cljs` and I fixed these.

Sorry for the space/format change if you may mind that, my Emacs automatically did that 😄  .

### DIFF
--- a/src/re_frame/interceptor.cljc
+++ b/src/re_frame/interceptor.cljc
@@ -1,7 +1,9 @@
 (ns re-frame.interceptor
   (:require
-    [re-frame.loggers :refer [console]]
-    [re-frame.interop :refer [ratom? empty-queue debug-enabled?]]))
+   [re-frame.loggers :refer [console]]
+   [re-frame.interop :refer [ratom? empty-queue debug-enabled?]]
+   [clojure.set :as set]
+   ))
 
 
 (def mandatory-interceptor-keys #{:id :after :before})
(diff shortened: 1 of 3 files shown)


src/re_frame/interceptor.cljc
```diff
@@ -1,7 +1,9 @@
 (ns re-frame.interceptor
   (:require
-    [re-frame.loggers :refer [console]]
-    [re-frame.interop :refer [ratom? empty-queue debug-enabled?]]))
+   [re-frame.loggers :refer [console]]
+   [re-frame.interop :refer [ratom? empty-queue debug-enabled?]]
+   [clojure.set :as set]
+   ))
```
thenonameguy: I think the closing parens should be at the end of the above line.

cmal: Ahh. sorry for that.

cmal: Excuse me, I am not quite sure about what you mean. Could you tell me the right way so that I can fix it. Thanks.

thenonameguy: It should be:
```clojure
[clojure.set :as set]))
```
instead of:
```clojure
[clojure.set :as set]
))
```

ruiyang: I think he means
it should be like the following at the end of the require.
 [clojure.set :as set]))

cmal: Sorry for that, I'll change this.

PR #348, opened 2017-05-31
fix requires to use the strict method

cmal:
I found re-frame use some clojure official libraries with the unstrict format, that make it impossible to be used with some tools like shadow-cljs and I fixed these.

Sorry for #347 I've closed that because being a different author at home. Fixed the typo of #347.


mike-thompson-day8:
Many Thanks.

PR #355, opened 2017-07-05
Remove extraneous code in todomvc sample

satchit8:
The event handler for :complete-all-toggle doesn't appear to require any info from the event vector, yet code that dispatches to it in the task-list component makes it look otherwise.


danielcompton:
Good catch, we're also doing that in the handler: https://github.com/satchit8/re-frame/blob/cba1acaddbbbf8ba4f17e93e1ac415fd305c94e4/examples/todomvc/src/todomvc/events.cljs#L128. I'm not sure which method we'd prefer here, I can see either being ok.

PR #359, opened 2017-07-13
Readme: fix what h should have returned

Grahack:
I know this is only a sketch. Maybe I'm wrong. I'm just trying to understand the cycle.

`db` should also have a value, but I understand that it's more complicated to show its value.

### DIFF
--- a/README.md
+++ b/README.md
@@ -338,7 +338,7 @@ An `effect handler` (function) actions the `effects` returned by `h`.
 
 Here's what `h` returned:
 ```clj
-{:db  (dissoc-in db [:items item-id])}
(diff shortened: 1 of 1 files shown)


Grahack:
OK for `db` as it is explained in note number 2 [here](https://github.com/Day8/re-frame/blob/master/docs/ApplicationState.md):

> In the documentation and code, I make a distinction between `app-db` (the `ratom`) and
> `db` which is the (map) `value` currently stored **inside** this `ratom`.

danielcompton:
Hmm, will defer to @mike-thompson-day8 on this one.

mike-thompson-day8:
Thanks!

PR #361, opened 2017-07-13
Fix typo in comments of simple/core.cljs

Grahack:


examples/simple/src/simple/core.cljs
```diff
@@ -13,7 +13,7 @@
     (rf/dispatch [:timer now])))  ;; <-- dispatch used
 
 ;; Call the dispatching function every second.
-;; `defonce` is like `def` but it ensures only instance is ever
+;; `defonce` is like `def` but it ensures only instance one is ever
```
danielcompton: "only one instance"?

Grahack:
It should be OK now, I added a commit. Sorry for the noise.

mike-thompson-day8:
And again, thanks.

PR #364, opened 2017-07-14
Simpler line in simple/core.cljs

Grahack:
It also makes the code like in the walkthrough:
<https://github.com/Day8/re-frame/blob/master/docs/CodeWalkthrough.md#reg-sub>


mike-thompson-day8:
Good to get this consistent, thanks.

PR #366, opened 2017-07-15
Interceptors doc: add "effect" to qualify handler

Grahack:
This change is only valid if I understood this correctly.
A handler in [this page](https://github.com/Day8/re-frame/blob/develop/docs/Interceptors.md) is an effect handler, right?
I'd also be happy if you say I understood correctly but you don't merge my PR!

BTW thanks for this fun and informative work on the docs.


Grahack:
Sorry, I messed up `event` and `effect`.

PR #378, opened 2017-07-16
App structure: add core.cljs for multipanel apps

Grahack:
I maybe wrong on this one.


mike-thompson-day8:
Good point!  Thanks for your detailed review and many suggestions..

PR #379, opened 2017-07-17
Fix typo in my-reg-event-fx function

pesterhazy:
The example code in the docs didn't work for me. This fixes the issue. Also remove trailing whitespace


pesterhazy:
Found another issue with the sample my-reg-* functions relating to duplicated interceptors - PR updated

mike-thompson-day8:
Thanks.  Sorry, that would have been annoying bug to find !!!

pesterhazy:
No worries, thanks for the quick merge :)

Issue #385, opened 2017-07-30
`re-frame.core/clear-subscription-cache!` doesn't work on the JVM


julienfantin:
The implementation of `clear-subscription-cache!` disposes of all the reactions in the cache, but  implicitly relies on the reactions' `on-dispose` hook to clear the map. 

Since on the jvm `dispose!` is a no-oop, that makes `clear-subscription-cache!` a no-op too.

julienfantin:
As an aside this can't be tested right now since the test suite only runs in ClojureScript. Is there any interest in converting the test suite to cljc, and if so should we do that before fixing that issue?

danielcompton:
Running re-frame on the JVM isn't something we do at the moment at Day8. However I'd be happy to look at a PR for converting the tests to CLJC, it would be best if it doesn't significantly add complication to running/writing CLJS tests. I'd also be open to a second PR to `interop.clj` which collects the `on-dispose` functions it gets given, and calls them when they are disposed.

PR #386, opened 2017-07-31
Add jvm interop handlers for dispose! and add-on-dispose!

julienfantin:
Added a couple of interop handlers to try and replicate reagent's disposal behavior on the JVM as per @danielcompton 's recommendation in #385. 

It was initially assumed that atoms did not need to implement this behavior since there are no real reactions and thus no retention cycles on the JVM, but the subscription cache relies on the `on-dispose` callbacks to clear the cache entry for that _reaction_. This behavior is also relied on in `subs/clear-subscription-cache!` which is essentially a no-op for the JVM on master now.

I also added a test for `subs/clear-subscription-cache!` which, although it doesn't actually test the code in this PR, should do so when the test suite is ported to cljc.

### DIFF
--- a/src/re_frame/interop.clj
+++ b/src/re_frame/interop.clj
@@ -23,6 +23,8 @@
 
 (defonce ^:private executor (Executors/newSingleThreadExecutor))
 
+(defonce ^:private on-dispose-callbacks (atom {}))
+
 (defn next-tick [f]
   (let [bound-f (bound-fn [& args] (apply f args))]
     (.execute ^Executor executor bound-f))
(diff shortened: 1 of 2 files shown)


julienfantin:
Submitted PR against the wrong branch, apologies.

julienfantin:
@danielcompton any feedback on why this didn't make it?

danielcompton:
Ah this was accidentally closed when we moved to everything being on `develop`. I thought I'd found everything but I missed this, will have a closer look. Please reopen it as we're still interested in this.

danielcompton:
Actually, I've just merged it in https://github.com/Day8/re-frame/commit/b7be7864a4d65b2eacafd18e4d9971490b8ddd37. Thanks and sorry for the delay.

PR #390, opened 2017-08-07
Change the require of reg-cofx

Grahack:
By the way, I don't understand why this call to `reg-cofx` belongs to `db.cljs`. Would it be really bad to put it in `events.cljs`?

Thanks.


mike-thompson-day8:
I can certainly see an argument for putting cofx into `events.cljs`.   I choose to home it in `db.cljs` because it felt data/state related.  Some might even argue it should go in a new `cofx.cljs` namespace.

mike-thompson-day8:
I prefer the explicit API namespace, so won't merge.

PR #391, opened 2017-08-07
Rename default-value to default-db

Grahack:
Some bits of the code are more readable with this name.


mike-thompson-day8:
Thanks

PR #394, opened 2017-08-08
Introduce app-global-interceptors

benedekfazekas:
Container and API to add application global interceptors to all events
without the need to define them on the `reg-event-` functions.

Adding interceptors an the `reg-event-` functions is still possible
for fine grained control.

`re-frame.core` is refactored to use the new API to add its own global
interceptors and also the interceptor section is moved to the top of
the namespace.

The example todomvc app make use of the new API to add the `debug`
interceptor to all event handlers.

`app-global-interceptors` is a map that contains two vectors under the
keywords `:head`  and `:tail`.

Interceptors in the head vector come before the event interceptors, interceptors in
the tail vector come after. Event interceptors in this context are the ones defined in
`reg-event-` functions.

This is a bit drafty but communicates my intentions I suppose. Let me know what you think.

If in favour I will add notes to the changelog and figure out how to add tests for this although not really sure about the latter.

### DIFF
--- a/examples/todomvc/src/todomvc/events.cljs
+++ b/examples/todomvc/src/todomvc/events.cljs
@@ -2,7 +2,7 @@
   (:require
     [todomvc.db    :refer [default-db todos->local-store]]
     [re-frame.core :refer [reg-event-db reg-event-fx inject-cofx path trim-v
-                           after debug]]
+                           after debug append-app-global-interceptor-tail]]
     [cljs.spec     :as s]))
 
 
@@ -58,9 +58,12 @@
 (def todo-interceptors [check-spec-interceptor               ;; ensure the spec is still valid  (after)
                         (path :todos)                        ;; 1st param to handler will be the value from this path within db
                         ->local-store                        ;; write todos to localstore  (after)
-                        (when ^boolean js/goog.DEBUG debug)  ;; look at the js browser console for debug logs
-                        trim-v])                             ;; removes first (event id) element from the event vec
+                        trim-v])                            ;;  removes first (event id) element from the event vec
 
+;; global config, setting global interceptors for all handlers
+;; look at the js browser console for debug logs
+(when ^boolean js/goog.DEBUG
+  (append-app-global-interceptor-tail debug))
 
 ;; -- Helpers -----------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


danielcompton:
Can you describe a little more about the motivations and use-cases behind this patch? I can see some, but want to understand more about what you're doing.

benedekfazekas:
yeah ofc, sorry for not being more elaborate on that.

at work where i started to use re-frame in anger spent days with a colleague trying to make sense of how a 3rd party lib generates re-frame events. would have been awesome to add the `debug` interceptor globally without the need to modify quite a few `reg-event`s across the codebase.

also as in general re-frame itself uses its own interceptors to handle events (`cofx/inject-db` for example). I think it makes sense to give more access (both read for visibility/inspection and write for control) to this machinery.

hope this makes sense

mike-thompson-day8:
I'm inclined to knock this PR back (which pains me greatly).  

Our position on global interceptors is explained in:
https://github.com/Day8/re-frame/blob/master/docs/FAQs/GlobalInterceptors.md

benedekfazekas:
Thanks for looking at this. Although I definitely want to make my case here, don't feel bad to say no as you are maintaining re-frame i should not and will not have any hard feelings ;)

Also sorry for missing the global interceptors doc page, my only excuse that it was created a few days before submitting my PR. Also after reading it -- hopefully this does sound very arrogant -- it feels more like a workaround for something perhaps multiple people want?

In my particular case described above that would mean again to refactor the whole project of mine to use this drop in replacement of `reg-event-db`. which is not a hard refactor to do but still quite excessive compared to a one liner to add a global interceptor.
Also on a newish project I would have to persuade my teammates to use my implementation of `reg-event-db` which I may or may not manage to do.

So if you have the time could you elaborate on why you would turn this PR down as the linked doc page does not really say why that would be a preferred way instead of opening up/adding an API to directly manipulate the existing interceptor chain/stack.

thanks again for you time to looking at this and also thanks for `re-frame`

mike-thompson-day8:
**On Timing*
I seem to remember that your question in clojurians was the trigger for me to write the FAQ page.  It wasn't the first time I've been asked and I'm on an FAQ missions ATM.  Which accounts for your PR and my FAQ happening around the same time. 

*On Reasons*
I'm trying to keep re-frame simple and minimal.  The goal is provide a base layer on which other layers can get built.  So deftly saying "no" to most new features is important.  And saying "no" to features which can be achieved via further layering is essential.

PR #395, opened 2017-08-09
Log warning when tracing not enabled

### LINKED ISSUE #384: Log a warning when user tries to add a trace callback but hasn't set a closure-define
`re-frame.trace/register-trace-cb` doesn't validate whether the closure define to enable tracing has been set or not. This could lead to confusing situations where people are trying to use re-frame's tracing abilities, but haven't set the closure-define so no traces show up.

The fix for this is to check at `re-frame.trace/register-trace-cb` whether tracing is enabled or not, and log a warning if it isn't, telling the user that they need to set the closure define for `re-frame.trace.trace-enabled?`.

saskali:
- problem: when people want to use tracing and forget to set closure-defines, traces don't show up and no errors or warnings are being displayed
- solution: check if tracing is enabled. if not, log warning message when opening trace panel  
- fixes #384

### DIFF
--- a/src/re_frame/trace.cljc
+++ b/src/re_frame/trace.cljc
@@ -27,7 +27,9 @@
   "Registers a tracing callback function which will receive a collection of one or more traces.
   Will replace an existing callback function if it shares the same key."
   [key f]
-  (swap! trace-cbs assoc key f))
+  (if trace-enabled?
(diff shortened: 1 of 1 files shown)


danielcompton:
Great stuff!

PR #397, opened 2017-08-14
Do not test if on-stop before calling it

Grahack:
The test may not be needed as we always call `todo-input` with an `on-stop` value.

I'm not 100% sure on this one, but I tested it.


danielcompton:
The current behaviour is intentional AFAICT. `todo-input` is a more generic input component, where sometimes you may not provide an `on-stop` callback function. Thanks for looking at this though, let me know if I've missed something here?

PR #405, opened 2017-08-31
fix some typos

jcw:
... and a big thank you for creating & supporting re-frame, as well as for this particular FAQ (still learning, awesome stuff!).


mike-thompson-day8:
Thanks

PR #409, opened 2017-09-06
Add console.dir to loggers

deg:


mike-thompson-day8:
@deg Can you explain the motivating usecase please?  re-frame doesn't currently use `console.dir`

deg:
One answer would be for symmetry and completeness.

But, my immediate motivation is that I was about to add something like this into the utils of one of my libraries that uses re-frame. But, it seemed silly to duplicate what you had, just to add in one more bit.
I’m happy using re-frame.loggers for my console needs, since it is already in my projects. Better than having to add, remember, or maintain another package.

mike-thompson-day8:
@deg I appreciate the thought and the effort, but I'm going going to merge this.  Less is better. No motivating usecase yet.

PR #413, opened 2017-09-12
Add dynamic var for warning on handler overwrite

### LINKED ISSUE #204: Don't warn about reloading handlers when initiated by figwheel

danielcompton:
When figwheel reloads code, it causes handlers to be re-registered which leads to a warning for each handler. This leads to noisy logs.  Figwheel has a :before-jsload key which can be called before it reloads application code.

This patch adds a `*warn-on-overwrite*` dynamic var, and shows how to use it in the example project.

One side effect of using this is that you will never get a warning for duplicate handlers when Figwheel is reloading. It may be possible to track how many handlers were reloaded in a particular Figwheel reload, but this would get very complex. In any case, you will still get warnings every time you refresh the browser, which should be good enough for the relatively rare case of creating duplicate handlers.

This still needs
- [ ] tests
- [ ] a decision on where to put the dynamic var (in `re-frame.core` or elsewhere?, if in core, what about circular dependencies?)
- [ ] Recommendations on the [usage of dynamic vars](https://groups.google.com/forum/#!topic/clojurescript/0FdQB4bwtSw)
- [x] Is `:before-jsload` supported/able to be used?
- [ ] Update changelog
- [ ] Docs on usage

Fixes #204

### DIFF
--- a/examples/todomvc/project.clj
+++ b/examples/todomvc/project.clj
@@ -17,7 +17,8 @@
                                                   :source-map           true
                                                   :source-map-timestamp true
                                                   :main                 "todomvc.core"}
-                                       :figwheel {:on-jsload "todomvc.core/main"}}}}}
+                                       :figwheel {:before-jsload "todomvc.core/before_reload"
+                                                  :on-jsload "todomvc.core/figwheel_reload"}}}}}
 
              :prod {:cljsbuild
                     {:builds {:client {:compiler {:optimizations :advanced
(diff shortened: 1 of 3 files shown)


danielcompton:
I've gotten a bit blocked by https://github.com/bhauman/lein-figwheel/issues/485. I think in the meantime we can release the method we've got above, and if a better option is introduced for something like `with-jsload` in the future, add support for it later.

mike-thompson-day8:
@danielcompton a reminder:  when discussing this recently, we realised that there are two distinct phases: 
1.  The period prior to first (figwheel ?) code reload  
2.  there after

We want warnings during 1, but never 2. 

So if we can catch the first (figwheel?) reload, then we can flip a switch so these warnings are not then reported.

superstructor:
Superseded. See #204. Thanks for the PR all the same 😄 Closing.

Issue #416, opened 2017-09-25
List of re-frame effect handlers / utils / projects


olieidel:
Thank you for the awesome work you put into re-frame, it's a great framework.

For some time now, I have been curating (well, that's a bit of an exaggeration) a list with re-frame resources, mainly effect handlers, utils and sample projects.

I think this could help quite a few people starting out with re-frame as other frameworks have similiar "awesome $framework" lists however for re-frame I haven't found anything like it. It's by no means complete, but it's a start.

So just a quick heads up - it's [here](https://github.com/olieidel/awesome-re-frame) if you want to link to it from somewhere (or just tell me where and I'll do the PR). If not, just close this issue :)

mike-thompson-day8:
@olieidel  we keep a list of external resources in here: 
https://github.com/Day8/re-frame/blob/master/docs/External-Resources.md

Would be great if you do a PR which added any new, useful links you have.

olieidel:
oh woops, I didn't know about that. Will check and create a PR if necessary.

PR #417, opened 2017-09-25
add links from awesome-re-frame

olieidel:
I also added `Debugging` and `React-Native` subcategories to `Tools, Techniques & Libraries`. Hopefully closes [#416](https://github.com/Day8/re-frame/issues/416).


danielcompton:
Thanks!

PR #418, opened 2017-09-25
Add subscription values to traces

saskali:
This PR adds subscription values to re-frame's traces. The value for `:sub/create` and `:sub-run` is being added to the associated tags of each subscription trace. 

- fixes day8/re-frame-trace#3

### DIFF
--- a/src/re_frame/subs.cljc
+++ b/src/re_frame/subs.cljc
@@ -114,6 +114,7 @@
      (if-let [cached (cache-lookup query)]
        (do
          (trace/merge-trace! {:tags {:cached?  true
+                                     :sub-value (deref cached)
                                      :reaction (reagent-id cached)}})
          cached)
 
@@ -123,7 +124,9 @@
          (if (nil? handler-fn)
            (do (trace/merge-trace! {:error true})
                (console :error (str "re-frame: no subscription handler registered for: \"" query-id "\". Returning a nil subscription.")))
-           (cache-and-return query [] (handler-fn app-db query)))))))
+           (let [sub-reaction (handler-fn app-db query)]
+             (do (trace/merge-trace! {:tags {:sub-value (deref sub-reaction)}})
+                 (cache-and-return query [] sub-reaction))))))))
 
   ([query dynv]
    (trace/with-trace {:operation (first-in-vector query)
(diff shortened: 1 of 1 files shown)


src/re_frame/subs.cljc
```diff
@@ -301,23 +304,30 @@
...
                reaction-id   (atom nil)
                reaction      (make-reaction
                                (fn []
-                                 (trace/with-trace {:operation (first-in-vector query-vec)
-                                                    :op-type   :sub/run
-                                                    :tags      {:query-v  query-vec
-                                                                :dyn-v    dyn-vec
-                                                                :reaction @reaction-id}}
-                                   (computation-fn (deref-input-signals subscriptions query-id) query-vec dyn-vec))))]
+                                 (let [sub-value (computation-fn (deref-input-signals subscriptions query-id) query-vec dyn-vec)]
```
danielcompton: The sub-value needs to be calculated within the `with-trace` so that we can capture accurate timings for it. I think you might need to use `trace/with-trace` to start the trace, calculate the sub value in a let, then use merge-trace to add it to the trace.

src/re_frame/subs.cljc
```diff
@@ -114,6 +114,7 @@
      (if-let [cached (cache-lookup query)]
        (do
          (trace/merge-trace! {:tags {:cached?  true
+                                     :sub-value (deref cached)
```
danielcompton: Not sure if this should be `:sub-value` or just `:value`?

src/re_frame/subs.cljc
saskali: that makes sense, I will move it so that it is calculated within the `with-trace`

src/re_frame/subs.cljc
```diff
@@ -123,7 +124,9 @@
          (if (nil? handler-fn)
            (do (trace/merge-trace! {:error true})
                (console :error (str "re-frame: no subscription handler registered for: \"" query-id "\". Returning a nil subscription.")))
-           (cache-and-return query [] (handler-fn app-db query)))))))
+           (let [sub-reaction (handler-fn app-db query)]
+             (do (trace/merge-trace! {:tags {:value (deref sub-reaction)}})
```
danielcompton: Hmm, I just realised that this will cause the subscription to run and be calculated immediately, rather than returning a reaction. I think it would be better to only capture the subscription value in `:sub/run`, not `:sub/create` as this could lead to (or mask) subtle bugs. Concretely, can you pull out the trace here and on line 116? We'll keep it in on the :sub/run. Sorry for not picking this up earlier.

danielcompton:
Released as https://clojars.org/re-frame/versions/0.10.2-beta1.

PR #429, opened 2017-11-26
Pretty list in the Full Stack FAQ

Grahack:
I just added stars to have bullets and the markdown is more consistent between the two last links.


danielcompton:
Thanks!

PR #431, opened 2017-12-01
todo-mvc initialise-db: avoid losing :showing set by router at page load time

kennytilton:
If we set the selection in todo-mvc to :done or :active then refresh the screen, the url will show the route correctly but show all todos. This is because the router sends the :set-showing event before the :initialise-db event, and the initialise-db event thinks it is indeed initialising the db and starts from default-db, missing that set-showing has already injected :showing, and we end up with ":showing :all" from default db.

More benignly, the db fails the spec validation in the :set-showing event at page load because the :todos key is missing.

Both issues have been addressed rather ham-handedly by always merging default-db and db in both events.


mike-thompson-day8:
Sorry for my slowness, and thanks for tracking this down, documenting it, and proposing a solution.  I'm going to use #452 as a base and modify, so will close this without merge.  But many thanks.

PR #432, opened 2017-12-01
Delete existing todo if saved blank, as per official

kennytilton:
The todo-mvc example has a serious deviation from the official spec which says existing todos, if edited to be blank, should be deleted. The app as is treats it as an NOP, leaving the todo unchanged.

This is not a bug, but the re-frame todo-mvc is offered as an exemplar (and the fix is trivial -- pass the value through to the two on-save handlers and adjusting them so they can implement their varying behaviors) so it might be worth conforming to the spec.


.idea/codeStyleSettings.xml
```diff
@@ -1,21 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
danielcompton: Can you keep this file please?

examples/todomvc/src/todomvc/views.cljs
```diff
@@ -1,15 +1,15 @@
 (ns todomvc.views
-  (:require [reagent.core  :as reagent]
```
danielcompton: Can you remove all the whitespace changes? It makes it hard to review.

kennytilton:
Not sure how to backout the whitespace changes so have redone this in new branch. Pull request on that coming up next.

PR #434, opened 2017-12-04
move seq test into individual save callbacks to differentiate handling of blanks

kennytilton:
The todo-mvc example has a serious deviation from the official spec which says existing todos, if edited to be blank, should be deleted. The app instead treats it as an NOP, leaving the todo unchanged.

This is not a bug, but the re-frame todo-mvc is offered as an exemplar (and the fix is trivial -- pass the value through to the two on-save handlers and adjusting them so they can implement their varying behaviors) so it might be worth conforming to the spec.


mike-thompson-day8:
I'm going to take your word for it and merge.  Hope you are right.  Many Thanks.

PR #436, opened 2017-12-06
Fix typo in sample code for 'reg-sub-raw' reaction fn

rojoangel:
Sample code in the `docs/Subscribing-To-External-Data.md` page has a typo that prevents the example from working.

A simple change makes the example functional:

```
(re-frame/reg-sub-raw
  :items
  (fn [app-db [_ type]]
       (let  [query-token (issue-items-query!
                            type
                            :on-success #(re-frame/dispatch [:write-to  [:some :path]]))]
         (reagent.ratom/make-reaction
           ;------------ the line below should be (fn [] (get-in @app-db [:some :path]))      
           (fn [] (get-in @app-db [:some :path] []))
           :on-dispose #(do (terminate-items-query! query-token)
                            (re-frame/dispatch [:cleanup [:some :path]]))))))
```


mike-thompson-day8:
Thanks!

PR #438, opened 2017-12-08
Add RealWorld to Examples

jacekschae:
Just pointing out that there is another reference to unfinished RealWorld example - point number 4.


danielcompton:
Thanks!

PR #439, opened 2017-12-21
Fix re-frame.interop/now for nodejs

kommen:
In a nodejs environment`js/performance` is not defined by default and would need `perf_hooks` to be required. See https://nodejs.org/api/all.html#perf_hooks_performance_timing_api

This is quick fix to make `re-frame.interop/now` not crash when it hasn't been required.

I'm not sure if there are any tests to be written as per the contributing guidelines for re-frame in a nodejs env?

### DIFF
--- a/src/re_frame/interop.cljs
+++ b/src/re_frame/interop.cljs
@@ -38,7 +38,9 @@
   (js/setTimeout f ms))
 
 (defn now []
-  (if (exists? js/performance.now)
+  (if (and
+       (exists? js/performance)
+       (exists? js/performance.now))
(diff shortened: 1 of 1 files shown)


danielcompton:
Thanks! We don't have any tests for node currently, so we don't need to add any here. @jeaye I think you've hit this isssue recently?

jeaye:
Yes indeed, I hit this with React Native. Looks like a good fix for that as well. Thank you!

danielcompton:
I've just deployed `0.10.3-beta` with this fix.

PR #451, opened 2018-02-16
DOC: Added "dispatch-n" to list of example effects in Effects.md

realgenekim:
I spent months searching for "re-frame multiple dispatch", but managed to somehow overlook `:dispatch-n` — until someone in Clojurian Slack channel pointed it out.  

I now see it referenced in source code and API docs.  But a little hint that it exists in this doc would have enabled me to find it right away.  Thank you!


mike-thompson-day8:
Thanks!

PR #452, opened 2018-02-17
Fix bugs in todomvc example

tonysherbondy:
- Routing events were able to be queued up and processed before initialize-db. (E.g., when you refresh  the page to http://localhost:3450/#/active)
- Passing invalid props to input. (Error messages in console)


tonysherbondy:
Possibly related to #431

mike-thompson-day8:
Thanks!

PR #459, opened 2018-03-13
Added sse-fx to fx/cofx list

milt:
Just adding [a little lib for sse we made](https://github.com/yetanalytics/sse-fx) to the external resources doc. Thanks again for the amazing lib!


mike-thompson-day8:
Thanks!

PR #460, opened 2018-03-14
Update CodeWalkthrough.md

cloojure:
- Fix step #5 to open local file instead of hitting http server.
 - Make file/directory references consistent (avoid appearance they are absolute pathnames).


cloojure:
OK, I see that was wrong. An "rm -rf *" and a clean clone fixed whatever was wrong.

PR #461, opened 2018-03-14
Update README.md

cloojure:
Fix step #5, just like step #4 in the CodeWalkthrough.md   file.


cloojure:
OK, I see that was wrong. An "rm -rf *" and a clean clone fixed whatever was wrong.

Issue #469, opened 2018-05-22
Reaching invalid state by dispatching continuosly


shakdwipeea:
I am trying to run re-frame on clojure. I am running into a invalid route when I dispatch continuosly.
The error msg I get is:
```
Exception in thread "pool-1-thread-1" clojure.lang.ExceptionInfo: re-frame: router state transition not found. :scheduled :finish-run {:fsm-state :scheduled, :trigger :finish-run}
```

I have created a repo to reproduce this:
https://github.com/shakdwipeea/reframe-state-transition-bug

danielcompton:
It looks like there may be a race condition here in the Clojure (note, not ClojureScript) code. I could reproduce this as well. After looking at the code for a few minutes, my best guess is that an event is dispatched while the executor is running and the FSM is in the `:running` state. The executor tries to run the `:finish-run` trigger, but underneath it, the FSM was moved back to the `:scheduled` state.

danielcompton:
Fixed in #471

piotr-yuxuan:
That feeling when I've been stumbling upon a weird error for some hours, finally I get convinced that nothing obvious in my code is responsible for it, I find this issue which has exactly the same symptoms and at the end of it I can read « Fixed » ❤️

piotr-yuxuan:
@scgilardi and @danielcompton thanks a lot for your work in https://github.com/Day8/re-frame/pull/471. However, might I suggest you to release it? ^^ looks like it's currently in master, but latest release comes from February.

PR #471, opened 2018-05-28
[#issue-469] -fsm-trigger depends on state consistency

scgilardi:
- re: Day8/re-frame#469
- `-fsm-trigger` reads from and writes to volatile state
- it works fine in cljs because that's single-threaded
- jvm clojure interop creates a single threaded executor, but that
  coexists with the main thread for a total thread count of 2
- adding `locking` on the `EventQueue` object avoids the race for
  clojure and is a no-op for clojurescript.

### DIFF
--- a/src/re_frame/router.cljc
+++ b/src/re_frame/router.cljc
@@ -126,46 +126,47 @@
     ;; Given a "trigger", and the existing FSM state, it computes the
     ;; new FSM state and the transition action (function).
 
-    (trace/with-trace {:op-type ::fsm-trigger}
-      (let [[new-fsm-state action-fn]
-            (case [fsm-state trigger]
+    (locking this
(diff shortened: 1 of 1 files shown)


danielcompton:
Thanks heaps for this, we really appreciate it (and your previous work on re-frame 😄).

scgilardi:
You're welcome! :)

And thanks for the test!

In running the test, I did notice that it needs a bit of a tweak. The test doesn't currently wait for the processing of the events to be complete before checking that they all got processed. Maybe add a "reached the end" (sentinel) event to the queue after the test events and write some code that detects when that one is processed (maybe an event callback), or maybe just poll the db (in the test) for a count on the array until it gets to 1000 and fail if that doesn't happen in a couple of seconds. Something idiomatic and effective. As it is now, the test passes only intermittently.

PR #474, opened 2018-05-31
update d3 reference to point to rid3 instead of blogpost

gadfly361:
Saw a link to a blogpost that predated my reagent + d3 library, rid3, and wanted to updated it


mike-thompson-day8:
Thanks!

PR #475, opened 2018-06-16
Prevent WARNINGs when using :infer-externs true + add deps.edn

galdolber:
WARNING: target/api/inferred_externs.js:40: WARNING - name goog is not defined in the externs.
goog.DEBUG;
^^^^

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -1,0 +1,6 @@
+{:deps
+ {org.clojure/clojure #:mvn{:version "1.8.0"},
+  org.clojure/clojurescript #:mvn{:version "1.9.908"},
+  reagent #:mvn{:version "0.7.0"},
+  net.cgrand/macrovich #:mvn{:version "0.2.0"},
+  org.clojure/tools.logging #:mvn{:version "0.3.1"}}}
--- a/deps.edn
+++ b/deps.edn
@@ -1,4 +1,5 @@
-{:deps
+{:paths ["src"]
+ :deps
  {org.clojure/clojure #:mvn{:version "1.8.0"},
   org.clojure/clojurescript #:mvn{:version "1.9.908"},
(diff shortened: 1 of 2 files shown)


danielcompton:
Can you explain a little bit more about what is happening here? Is the `deps.edn` file necessary, or just a debugging convenience for you? What were you running that gave you that warning?

galdolber:
The deps.edn file allows using re-frame straight from git with the [clojure cli](https://clojure.org/guides/deps_and_cli).

The other fix is to avoid a warning when using :infer-externs. The warning is actually an error on the generated externs. See [https://github.com/clojure/clojurescript/commit/6048bc95001e1d5f1c996c5cda187cdfc0193bc8](https://github.com/clojure/clojurescript/commit/6048bc95001e1d5f1c996c5cda187cdfc0193bc8)

danielcompton:
Ok, we'll look at the two changes separately, can you explain more about the impact of removing the `js/`. Will this work on older versions of ClojureScript?

galdolber:
Sounds good. Yes, this is backward compatible.

superstructor:
Could you please split this into two changes @galdolber ? 

Removing the `js/` is pretty straight forward to merge.

The `deps.edn` is less compelling. I am not convinced we need it. For example, you can still use re-frame releases from clojars using the clojure cli. Is this intended so you can depend on unreleased versions ? If we did support that, then we would have to do it properly by introducing deps.edn plugin for lein so that dependencies were not duplicated between `deps.edn` and `project.clj`.

superstructor:
Infer warning for goog.DEBUG fixed already on master in [467b3f3](https://github.com/day8/re-frame/commit/467b3f3dd4e4dde2246c3bc846b8d6eda3eae3ca) and we already have a `deps.edn` file so closing. Thanks for the PR regardless.

PR #487, opened 2018-09-02
Docs: Small readme update for devs coming from javascript

lane-s:
This is a miniscule update, but I think it may be helpful to people coming from javascript. I may be wrong about this, but I think 'Domino 4' is very similar to `mapStateToProps` in Redux.


mike-thompson-day8:
Thanks

PR #489, opened 2018-09-12
Clarify passing of props

bmaddy:
This threw me for a while and I see that it also confused Zach, whose blog post is cited as an example
http://zachcp.org/blog/2015/reagent-d3/ (search for `reagent/props` to see his discussion about his confusion)


mike-thompson-day8:
Thanks!

PR #490, opened 2018-09-20
upgrade versions to work with Java 9 & 10

cloojure:
small changes to `project.clj`


danielcompton:
The diff on this is pretty massive, are you able to open a branch which contains just what is needed for Java 9/10?

PR #495, opened 2018-09-23
fix docstring formatting of code snippets

martinklepsch:
Many docstrings in re-frame use Markdown but don't properly indent code snippets by 4 spaces resulting in broken formatting [on cljdoc](https://cljdoc.xyz/d/re-frame/re-frame/0.10.5/api/re-frame.std-interceptors). This PR fixes this for the `std-interceptors` namespace.

via https://github.com/cljdoc/cljdoc/pull/117


danielcompton:
Pushed on master, not sure why it didn't close this one.

PR #509, opened 2018-10-22
small correction

cloojure:


mike-thompson-day8:
I always appreciate input but I'm going to leave the existing code in place and stay with using `console.debug`

PR #511, opened 2018-10-28
Avoid running after/enrich without a db effect

### LINKED ISSUE #453: Only run after and enrich interceptors against an effect db, not coeffect db
Following on from #447 and #450, it was discovered that there is probably not a lot of point in running `after` or `enrich` interceptors against a coeffect db if an effect db wasn't returned. This is technically a breaking API change, but I'd be surprised if it affected anyone.

### Current behaviour

If an event handler doesn't return a db effect, an `after` or `enrich` interceptor is run against the db _coeffect_.

### Proposed behaviour

If an event handler doesn't return a db effect, no `after` or `enrich` interceptor is run.

frenata:
Previously both standard interceptors acted upon the `db` *coeffect* if
no *effect* was available. Per #453 and related discussion, there is little
reason in general to do this, since this context means that the `app-db`
should not change.

Both functions are changed to explicitly check for the presence of the
`db` effect. If it is not present in the context, the interceptor
function is not run.

The previous functionality around `nil` and `false` values for the `db`
effect is preserved: the interceptors *are* run in this case. Tests are
added where needed to verify the behavior of both functions in the case 
of non-present, falsey and truthy values.

The relevant docstrings are tweaked to accommodate this change.

Closes #453.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,8 @@
 ## Unchanged
 
+#### Fixed
+
+  - `after` and `enrich` interceptors now no longer run if there is no `db` effect, rather than running against the `db` coffect. [#453](https://github.com/Day8/re-frame/issues/453)
 
 
 ## 0.10.6 (2018-09-03)
--- a/src/re_frame/std_interceptors.cljc
+++ b/src/re_frame/std_interceptors.cljc
@@ -240,7 +240,7 @@
   return a modified `db`.
 
   Unlike the `after` interceptor which is only about side effects, `enrich`
-  expects `f` to process and alter the given `db` coeffect in some useful way,
+  expects `f` to process and alter the given `db` effect in some useful way,
   contributing to the derived data, flowing vibe.
 
   Example Use:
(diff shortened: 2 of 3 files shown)


superstructor:
Having read the prior discussion in #453, #447 and #450 and reviewed the code I am satisfied this change implements an agreed change in behaviour so merging. Sorry it took awhile to get to this and thanks for the PR @frenata !

mike-thompson-day8:
We need to back this out.  This is a breaking change.

mike-thompson-day8:
@frenata 
I'm really sorry we had to back this out after you put the effort in, and got this PR perfect.  
It is my fault that this has happened.  I should have recognised the problems with this change before you went to the effort.  I just didn't think through the implications of the breakage correctly. 

The best approach from here is to leave the existing interceptors as is, and to add two *new* interceptors which have the alternative behaviour.  There is a case to both approaches.

frenata:
@mike-thompson-day8 No big deal, these things happen!

I'm happy to help with further work in this area if that would be helpful, but otherwise am happy to have had a chance to dig around in the re-frame internals. It's a great project that I was happy to have used at work and I was interested in giving back in some small way.

superstructor:
This has been reverted in [0.10.8](https://github.com/Day8/re-frame/blob/master/CHANGES.md#0108-2019-07-15). Sorry for the inconvenience. I misjudged the change as low risk.

PR #513, opened 2018-11-08
update number of SLOC

AndreaCrotti:
I was surprised it was only 500 and just checking with `cloc` it looks like it's more around 1500 now

### DIFF
--- a/docs/FAQs/DoINeedReFrame.md
+++ b/docs/FAQs/DoINeedReFrame.md
@@ -67,7 +67,7 @@ another you'll be using "Reagent + a broader architecture".
 
 Hmm. I feel like I'm missing a few, but that's certainly an indicative list.
 
-re-frame is only about 500 lines of code.  So it's value is much more in the honed 
+re-frame is only about 1500 lines of code.  So it's value is much more in the honed
 choices it embodies (and documents), than the code it provides.
 
 ### What Reagent Does Provide


mike-thompson-day8:
Haven't checked line counts in a long time, but I'm guessing the 1500 number includes comments?

AndreaCrotti:
This is what I get with `cloc` just against src
```
$ cloc src/
      15 text files.
      15 unique files.                              
       0 files ignored.

github.com/AlDanial/cloc v 1.80  T=0.01 s (1694.5 files/s, 247740.5 lines/s)
-------------------------------------------------------------------------------
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
ClojureC                        13            373            273           1398
Clojure                          1             18             21             52
ClojureScript                    1             14              4             40
-------------------------------------------------------------------------------
SUM:                            15            405            298           1490
-------------------------------------------------------------------------------
```

martinklepsch:
It should be noted that `(comment ,,,)` forms count as source code as far `cloc` is concerned.

mike-thompson-day8:
Okay, I just did a quick, inaccurate audit. 

The first file I just looked in `fx.cljc` and it contains about 60-ish  lines of code despite being 171 lines long.  The rest is docstrings, comments and blanks. 
https://github.com/Day8/re-frame/blob/master/src/re_frame/fx.cljc

Next file I looked into was  `std_interceptors.cljc` which looks to be about 50% docstrings:
https://github.com/Day8/re-frame/blob/master/src/re_frame/std_interceptors.cljc

I then  `db.cljc` but it is unusual so I ignored it (three line of code out of 10). 

At that point, I stopped.   In the context of saying "there's not much code in re-frame really", it still feels to me that a round number like 500 loc  is more a more accurate indication of complexity than 1500, even if  650 is more accurate these days.

mike-thompson-day8:
I'll accept this PR because the number should be updated, but I'll probably then re-edit it to 750 to be on the safe side.

mike-thompson-day8:
Thanks!

AndreaCrotti:
Ah ok I see, the only problem with `cloc` as far as I can tell is that it counts docstrings also has code, which is what's skewing up the total number.
Ok thanks!

martinklepsch:
Was curious if there were any tools that would count lines properly for Clojure, and here we go: [Counting Clojure Code](http://aaroniba.net/counting-clojure-code) (which introduces [lein-count](https://github.com/aiba/lein-count)).

AndreaCrotti:
Ah nice @martinklepsch , I didn't find that because I was just trying the generic SLOC counters, and apparently none of the one I tried really work well with Clojure.
This is definitively going to be more accurate, thanks

Issue #512, opened 2018-11-04
Error with infer-externs "goog is not defined in the externs"


olivergeorge:
I've turned on :infer-externs and now get a warning doing an advanced optimisations build:

```
WARNING: target/ios/inferred_externs.js:1: WARNING - name goog is not defined in the externs.
goog.DEBUG;
^^^^
```

Pretty sure this is caused by the way goog.DEBUG is referred to in the re-frame.interop namespace

```
(def ^boolean debug-enabled? "@define {boolean}" ^boolean js/goog.DEBUG)
```

My testing of infer-externs shows this:

```
; Good
(when ^boolean goog.DEBUG (println "Hello world!"))
(when ^boolean goog/DEBUG (println "Hello world!"))

; Bad
;(when js/goog.DEBUG (println "Hello world!"))
```

danielcompton:
Just to clarify, you're saying it looks like the issue is the leading `js/` on `goog.DEBUG`?

olivergeorge:
Yes, seems to be.
```
$ cat deps.edn 
{:deps {org.clojure/clojurescript {:mvn/version "1.10.439"}}}

$ cat compile-opts.edn 
{:infer-externs   true
 :optimizations      :advanced }

$ cat src/can_i_infer/core.cljs 
(ns can-i-infer.core)

(when js/goog.DEBUG (println "Hello world!"))

$ rm -r out/

$ clj -m cljs.main -co compile-opts.edn -c can-i-infer.core
Nov 05, 2018 3:01:58 PM com.google.javascript.jscomp.LoggerErrorManager println
WARNING: out/inferred_externs.js:9: WARNING - name goog is not defined in the externs.
goog.DEBUG;
^^^^

Nov 05, 2018 3:01:58 PM com.google.javascript.jscomp.LoggerErrorManager printSummary
WARNING: 0 error(s), 1 warning(s)
```

mike-thompson-day8:
@olivergeorge any chance you could provide a PR for the fix you see as working?

mike-thompson-day8:
Fixed via https://github.com/Day8/re-frame/pull/517

PR #517, opened 2018-11-25
Avoid warning when using :infer-externs

olivergeorge:
Ref https://github.com/Day8/re-frame/issues/512


mike-thompson-day8:
Thanks!  Fixes #512

PR #521, opened 2019-01-10
Edit README.md

hgdsraj:
There are a lot of sentences in the first bit that start with "And" (unnecessarily). Removed the 3rd "and" start. AND the fourth.

Changed "I cannot stress enough what a big deal this is" to "The importance of this cannot be stressed enough" but I'd like a further change since I don't actually understand what "this" is fully in that context.

Changed "etcetera, boom, boom, boom" to "ad infinitum" 

There are far to many sentences starting with And. This should be changed in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ McCoy might report "It's MVC, Jim, but not as we know it".  And you would respon
 re-frame is a **functional framework**."
 
 Being a functional framework, it is about two things: data, and the functions 
-which transform that data.  And, because it is a reactive framework, the ["data coordinates the functions"](https://www.youtube.com/watch?v=ZgqFlowyfTA&t=80) (and not the other way around).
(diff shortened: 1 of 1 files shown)


mike-thompson-day8:
Many Thaks. I really DO appreciate all attempts to improve the library and its documentation, but in this particular case, we'll stick with the idiosyncratic style we have.  It is a communication style which has served us well so far.

mike-thompson-day8:
Many Thaks. I DO appreciate all attempts to improve the library and its documentation, but in this particular case, I'll stick with my idiosyncratic style in these places. It is a communication style which has served me well so far.

BTW, I note that today alone you have created 23 PRs against different GitHub READMEs.  23 ... in one day. So it almost seems like you are running a grammar-bot over these GitHub READMEs and creating PRs. from it? That may find a few problems here and there, but it will also tend to eliminate any character in "the author's voice", which can be useful for creating reader engagement with otherwise dry slabs of technical words.

hgdsraj:
Thanks for your response :) 
I'm not running a grammar bot, just spending some time contributing to various READMEs for a project for a university class on technical documentation and technical writing.

Issue #486, opened 2018-09-01
Namespaced keywords


AndreaCrotti:
In one project we use namespaced keywords to register all the events, which has the advantage that you get autocompletion and you don't misspell a keyword as easily.

I know it's not really necessary, but I wonder if it would be worth to mention that in the documentation, or even use that as default style for examples and co.
Any thoughts about that? Or non namespaced keywords are preferred for some reason?

mike-thompson-day8:
Happy to take a PR if you find the right place and words.

lwhorton:
Just some food for thought:

I've developed a handful of rather large reframe projects at this point. Originally I tried to use events to model my domain in a way that I could, if necessary, swap out view implementations (as the needs of the UI changed) while keeping my events/subscriptions the same. It seemed to me a necessary requirement that an effective model would therefore *not* include namespacing, and every event would be descriptive enough to stay "top level". Either I'm not disciplined enough, am incapable of correctly modeling my domain, or this is just not a feasible approach.

Time and time again I find that my domains are often too large to model via events without eventually falling back to namespaced events. Once that possibility opens up, though, it becomes easier to separate "feature specific" events from "domain events". Also, I find that namespacing domain events allows me to leverage different context in which an event can take place. Similar to bounded contexts in domain driven design, my events start to model similar-but-different things that take place in various contexts. In concrete terms: `:admin/change-user-email` has a very different flow than `:user/change-user-email`. 

I'd love to hear someone else's approach to solving this problem.

AndreaCrotti:
Yeah I agree @lwhorton I don't think it's a feasible approach to have everything top level, and it doesn't really make sense to do so imho.

In my experience everything bigger than the TODO mvc app, or any SPA with at least 2 pages, should try to split things by domain, and not by events/subscriptions/etc.
So well I just opened this issue because imho suggesting this default might lead people in the wrong direction, even if of course in the other hand when you are just getting started this model still makes sense.

I'm not too sure where we could add this suggestion though, maybe another document title "beyond the basic App" or something like that?

AndreaCrotti:
Actually sorry it's kind of already mentioned here
https://github.com/Day8/re-frame/blob/master/docs/Basic-App-Structure.md#larger-apps

even if it's the `Basic-App-Structure` file which is a bit confusing.
So I could just add something about namespaced keywords there and maybe rename that file to something more generic, any thoughts @mike-thompson-day8 ?

mike-thompson-day8:
[...]

PR #528, opened 2019-02-13
move the basic app structure file and add docs about namespaced keywords

AndreaCrotti:
This PR is the result of this discussion:
https://github.com/Day8/re-frame/issues/486


AndreaCrotti:
Sorry for all the extra whitespace deleted, I had Emacs set to delete trailing whitespaces, I can revert those changes if it's a problem.

PR #531, opened 2019-03-22
improve convenience for deps.edn contributors

raymcdermott:
If you are trying to hack on `re-frame` and your project is based around `tool.deps` it's not possible to use a local-root without having a `deps.edn` file in the re-frame project.

It's not a lot of work to make that happen but this would make the process slightly more efficient.

As an added value, I would also be happy to set the project up with [lein-tools-deps](https://github.com/RickMoynihan/lein-tools-deps) if that would be interesting to the project maintainers. This approach offers a nice balance between maintaining dependencies using the new system and retaining all of the `lein` tooling that the project currently enjoys.

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -1,0 +1,17 @@
+
+;; Allows users of deps.edn to more conveniently fork / make PRs to re-frame
+
+{:deps {org.clojure/clojure       {:mvn/version "1.8.0"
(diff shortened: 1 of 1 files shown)


danielcompton:
Thanks! Will leave lein-tools-deps for now, but maybe in the future.

Issue #522, opened 2019-01-14
Calling Subscriptions from Clojure REPL


Wolfgang51:
After doing some research I found that most of re-frame is now in `.cljc` (https://github.com/Day8/re-frame/blob/master/docs/FAQs/Why-CLJC.md). This was excited to hear this due to the fact that we can now test from the JVM now and not have to rely on a JV environment. However, when I tried exercising _subscriptions_ in the REPL I came to the following issues.

**1. Calling subscriptions that subscribe to other subscriptions and use their values for computations lead to an error message of invalid arity.** 

--- Subscription ---
```clj
; Subscription
(re-frame/reg-sub
 ::suggestions
 (fn [db [_ query]]
   (get-in db [:typeahead :suggestions])))

(re-frame/reg-sub
 ::query
 (fn [db [_ query]]
   (get-in db [:typeahead :query])))

(re-frame/reg-sub
 ::typeahead
 (fn [_ _]
   [(re-frame/subscribe [::query])
    (re-frame/subscribe [::suggestions])])
 (fn [[query suggestions] _]
   {:query query
    :suggestions suggestions}))
``` 
-- Error --
```
slm.components.typeahead.subscriptions=> (re-frame/subscribe [::typeahead])

Execution error (ArityException) at re-frame.subs/reg-sub$subs-handler-fn (subs.cljc:325).
Wrong number of args (1) passed to: slm.components.typeahead.subscriptions/eval28820/fn--28821

(let [typeahead @(re-frame/subscribe [::typeahead]                                         )]
                                    #_=>   typeahead)

Execution error (ArityException) at re-frame.subs/reg-sub$subs-handler-fn (subs.cljc:325).
Wrong number of args (1) passed to: slm.components.typeahead.subscriptions/eval27322/fn--27323
```

**2. Can't extract values from subscriptions** 

```clj
; db
(def default-db
  {:name "olanext"})

; subscription
(re-frame/reg-sub
 ::name
 (fn [db _]
   (:name db)))
```
--- attempt to extract value ---
```
slm.components.typeahead.subscriptions=> (let [name @(re-frame/subscribe [::name])]
                                    #_=>   name)
nil
```

Is there any resolution to these issues?

dawran6:
> 2. Can't extract values from subscriptions

I think this is because you haven't put any data into the app-db yet. You can do this by running this in the REPL:

```clojure
(re-frame/reg-event-db
 :initialize-db
 (fn [_ _] default-db))

[...]

PR #541, opened 2019-08-15
Support for layer 3 subscriptions in clj environment

dawran6:
This PR is to address issue #522.
The PR implements one of the proposals in the [comment](https://github.com/Day8/re-frame/issues/522#issuecomment-521764881).

Thank you for reviewing. Please let me know what you think.


superstructor:
Thanks for the fix @dawran6 👍  Merged.

PR #542, opened 2019-08-20
Migrate to shadow-cljs

superstructor:
Migrates the build of re-frame itself to [shadow-cljs](http://shadow-cljs.org/).

Also better supports consumers of re-frame using shadow-cljs via inclusion of `deps.edn`.

Should not break consumers of re-frame using any other build system, including figwheel.

Uses the recently released [lein-shadow](https://gitlab.com/nikperic/lein-shadow) `0.1.5` to manage `shadow-cljs.edn` and `package.json`.


mike-thompson-day8:
@superstructor should we also change across the `/example` apps?

PR #546, opened 2019-09-08
Clarify debug interceptor log message

mbertheau:
The debug interceptor logs the changes to the app-db that resulted from handling
an event. There are two cases without changes to the app-db: one, the effects
resulting from the event don't contain a `:db` key, and two, there is a `:db`
key in the effects, but the app-db is the same as before the event.

These two cases had a similar but slightly different log message. However, for
the uninitiated it was practically impossible to determine which of the two
cases described above is given based on the log message.

Improve that by talking about "no changes in" for the first case and about "no
changes resulted from" in the second case.


superstructor:
Thanks @mbertheau this is a good improvement and my apologies for the long delay in me getting to your PRs!

PR #547, opened 2019-09-08
Extract event from coeffects in the same place as in db-handler->inte…

mbertheau:
…rceptor

A good principle to follow in designing anything is to make things that are the
same look the same, and things that behave differently look differently.

`db-handler->interceptor` and `fx-handler->interceptor` are very similar. Their
only difference is that the latter calls the `handler-fn` with all coeffects and
assoces the result to the `:effects` key directly, whereas the former calls the
`handler-fn` with just the db from inside the coeffects, and assoces the result
to just the db in the effects.

However the code of these two functions slightly differed in *where* the event
and the db resp. the whole coeffect was extracted from the context. Eliminate
that difference so that it is clearer what's different and what's the same about
these two functions.

### DIFF
--- a/src/re_frame/std_interceptors.cljc
+++ b/src/re_frame/std_interceptors.cljc
@@ -140,13 +140,13 @@
   :id     :fx-handler
   :before (fn fx-handler-before
             [context]
-            (let [{:keys [event] :as coeffects} (:coeffects context)
-                  new-context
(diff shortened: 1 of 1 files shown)


superstructor:
Thanks @mbertheau and apologies for the delay in getting to this.

PR #548, opened 2019-09-09
Consistently use (co)effect helpers

mbertheau:
(Note: this builds on https://github.com/Day8/re-frame/pull/547)

In `re-frame.interceptor` there are a few functions that help you modify the
context during the processing of an event. In `re-frame.std-interceptors` these
helpers were used in some but not in all places that they can be used.

This PR changes the code so that `std-interceptors` uses these helpers where
possible.

I'm not sure these helpers are really helpful, for two reasons:

1. They are very thin wrappers around accessing the `:effects` resp.
`:coeffects` key in a context. As you can see in the diff of this pr, the
difference to using the standard clojure functions to modify a context is not
very pronounced.
2. They don't cover the following use-cases, that are taken from
`re-frame.std-interceptors`:

- dissocing a key from the coeffect. This is used in `trim-v` to restore the
  context to its original form in its `after` function. Note that the `path`
  interceptor is content with leaving its `:re-frame-path/db-store` key in the
  context.
- associng a whole effects map to the context. This is used in
  `fx-handler->interceptor` to replace the `:effects` in the context with the
  result of the call to the handler function. `db-handler->interceptor` could
  make use of the helper `assoc-effect` because it assoces just the `:db` key of
  the effect in the context.

In light of this I'd argue that the helpers don't pull their weight and
shouldn't be used.

If you agree, I can prepare a PR to that end.

Since the helpers are part of the public API, they would probably have to
remain, but their usage in `std-interceptors` can be removed.

### DIFF
--- a/src/re_frame/std_interceptors.cljc
+++ b/src/re_frame/std_interceptors.cljc
@@ -1,7 +1,7 @@
 (ns re-frame.std-interceptors
   "contains re-frame supplied, standard interceptors"
   (:require
-    [re-frame.interceptor :refer [->interceptor get-effect get-coeffect assoc-coeffect assoc-effect]]
+    [re-frame.interceptor :refer [->interceptor get-effect get-coeffect assoc-coeffect assoc-effect update-coeffect]]
     [re-frame.loggers :refer [console]]
     [re-frame.registrar :as registrar]
     [re-frame.db :refer [app-db]]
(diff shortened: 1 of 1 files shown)


superstructor:
I agree that such thin wrappers don't have much value. I personally prefer vanilla Clojure in such cases as its more transparent exactly what is occurring, and the wrappers are not any shorter/simpler.

superstructor:
Despite saying earlier that I agree these have doubtful value. I guess its also not a bad thing to maintain consistency with a publicly provided API so that at least it can aid understanding in how re-frame works, or how those API fns might be used. Merged.

PR #549, opened 2019-09-12
Improve hot reloading in example apps

simon-katz:
This makes hot reloading of the example apps change the UI without the need for a browser refresh.

(EDIT: Deleted words about text displayed before the app is loaded.)

### DIFF
--- a/examples/simple/src/simple/core.cljs
+++ b/examples/simple/src/simple/core.cljs
@@ -81,8 +81,21 @@
 
 ;; -- Entry Point -------------------------------------------------------------
 
+(defn render
+  []
+  (reagent/render [ui]
+                  (js/document.getElementById "app")))
+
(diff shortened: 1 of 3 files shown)


simon-katz:
I've just seen that it's possible to set up the after-load thing using metadata on the function definition rather than with an `:after-load` entry in the build config.

It would look like `(defn ^:dev/after-load render …)`.

See https://shadow-cljs.github.io/docs/UsersGuide.html#_metadata

If you'd prefer it that way I can make the changes.

simon-katz:
I wanted to flush this out of my head, so I've made the following changes:

- A single commit.
- Use `:dev/after-load` metadata on the function that's called after hot-reloading.
- Only have changes relating to hot-reloading.

superstructor:
Thanks for the improvements @simon-katz !

PR #550, opened 2019-09-16
Improve the text that is displayed before app load in example apps

simon-katz:
In the simple example, say "Reframe" instead of "Reagent" in the that is displayed before app load.

In the todomvc app, add text that is displayed before app load.


mike-thompson-day8:
Thanks!

PR #551, opened 2019-09-20
Fix typos in example after-load function comments

simon-katz:
Sorry — just spotted some typos in my previous PR.


mike-thompson-day8:
Thanks

PR #557, opened 2019-10-18
Update simple/CodeWalkthrough.md

mack1070101:
`lein do clean, figwheel` no longer works, but `lein do clean, shadow watch client` does. Additionally, update link to http://localhost:8280/example.html as suggested by simple/README.md


superstructor:
Thanks @mack1070101 for improving the docs :smile: :+1:

PR #562, opened 2019-11-16
Add debug log level

### LINKED ISSUE #561: Debug log level is missing
`js/console.debug` is a thing: https://developer.mozilla.org/en-US/docs/Web/API/Console/debug
And it's well supported: https://caniuse.com/#feat=mdn-api_console_debug

Alas, it's impossible to add if from a user's side because of an assert on `(contains? @loggers level)`.

p-himik:
Fixes #561


superstructor:
Thanks @p-himik !

Issue #564, opened 2020-01-06
Add docs - avoid subscribe from JS event handlers


smahood:
Based on a conversation in Slack, we should add something to the docs to avoid using `subscribe` inside JS event handlers such as `:on-click`. 
I would lean towards a separate FAQ entry, but it is very similar to [How can I use a subscription in an Event Handler](https://github.com/day8/re-frame/blob/master/docs/FAQs/UseASubscriptionInAnEventHandler.md) and could probably be added there instead.

If nobody else beats me to it, I'll try to write it up a little later this week.

PR #565, opened 2020-01-10
Add FAQ for using a subscription in a JS event

smahood:
Initial attempt at docs for issue #564

### DIFF
--- a/docs/FAQs/UseASubscriptionInAJsEvent.md
+++ b/docs/FAQs/UseASubscriptionInAJsEvent.md
@@ -1,0 +1,53 @@
+### Question
+
+How do I access the value of a subscription from within a Reagent or JavaScript event (such as on-click)?
+
+### The Wrong Way
+
+You should NOT do this:
+```clj
+[:input {:type "button" 
+         :value "Click me!"
+         :on-click #(let [sub-val  @(subscribe [:something])]   ;;<--- Eeek
+       ....)}]
+```
+
+because that `subscribe`:
+1. might create a memory leak (the subscription might not be "freed")
+2. makes the event handler impure (it grabs a global value)
+
+
+### The Better Way
+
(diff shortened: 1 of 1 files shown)


docs/FAQs/UseASubscriptionInAJsEvent.md
```diff
@@ -0,0 +1,53 @@
...
+```clj
+[:input {:type "button" 
+         :value "Click me!"
+         :on-click #(let [sub-val  @(subscribe [:something])]   ;;<--- Eeek
+       ....)}]
+```
+
+because that `subscribe`:
+1. might create a memory leak (the subscription might not be "freed")
+2. makes the event handler impure (it grabs a global value)
```
p-himik: I would remove this point.
Such event handlers are almost always anonymous functions created and used right then and there. And if you put `@` outside in `let`, it won't make the value "non-global".
I think creating a closure over a locally referenced value is just fine in this particular case, no matter where the data came from.

docs/FAQs/UseASubscriptionInAJsEvent.md
```diff
@@ -0,0 +1,53 @@
...
+
+
+### The Better Way
+
+If your `on-click` function requires access to re-frame data, the first thing you should do is convert it to a re-frame event and dispatch that instead.
```
p-himik: Not necessarily. It depends on the intention.
E.g. it can be a mostly Reagent component that just needs some data. It doesn't want to affect any outside state, it has its own. In that case, creating an separate re-frame event many make no sense.
I would say here that it's required to deref all subs during the render. And if a user needs a sub value in a JS event, they must create the event handler for each new value of the sub. And the solution presented here is just another way to do that for some of the cases.

superstructor:
Although @p-himik has made some suggestions worth following up, the existing changes have value in of themselves and any further improvements can be via a new PR from anyone who is available to make the changes. 

Thanks @smahood @p-himik 👍🏻 😄

PR #573, opened 2020-02-11
Adds missing "map" in PollADatabaseEvery60.md

Jeel-Shah:
## Core
For the defensive portion of `PollADatabaseEvery60.md`, `interval-handler` contains a `doseq`. Upon trying out the code, there is compilation problem and this is because `doseq` requires a seq of expressions. Earlier, it was missing this sequence presumably because map was missing and so no sequence was created and it failed to compile. I've added `map` and this resolves the issue.


superstructor:
Thanks for the fix.

I think in this case `run!` or `doall` would be more idiomatic. 

We don't need any bindings (hence the `_` in the vector) which is confusing to read, so we may as well use a form that doesn't provide bindings  ?

Jeel-Shah:
Yes, you're right! I've made the change.

superstructor:
Thanks @Jeel-Shah :+1:

PR #577, opened 2020-04-02
Change out deprecated reagent/render for reagent.dom/render

pathammer:
reagent/render got marked deprecated [recently](https://github.com/reagent-project/reagent/commit/a8c401859d1ed4de026117c08807813aedb81f16) and causes warnings in re-frame sample applications.


pathammer:
FYI @mike-thompson-day8

superstructor:
Thanks for the fix @pathammer Has been made to some other projects already (e.g. re-frame-template, re-frame-10x etc) so good to catch this one that had been missed.

PR #578, opened 2020-04-05
Consistent advice on naming of event handlers file

rlovtangen:
Example was based on the event handlers file being named 'handlers.cljs' right after being advised to name the event handlers file 'events.cljs'


superstructor:
Thanks for the fix @rlovtangen 👍 Your fix has been included in an upcoming major rebuild of the documentation but I can't merge it here as it could cause merge conflicts with that work.

PR #583, opened 2020-04-17
chore: replace deprecated `reagent.core/render`

aviad:
see https://github.com/reagent-project/reagent/blob/master/CHANGELOG.md#0100-2020-03-06
replaced with `reagent.dom/render` in the examples


superstructor:
Duplicate of #577 that was just merged. Thanks for the PR regardless @aviad

PR #584, opened 2020-04-19
Update Interceptors.md

coxy1989:


superstructor:
Thanks for the fix @coxy1989 👍 Your fix has been included in an upcoming major rebuild of the documentation but I can't merge it here as it could cause merge conflicts with that work.

PR #585, opened 2020-04-23
Update core.cljs to use reagent.dom/render

luposlip:


superstructor:
Duplicate of #577 that was just merged. Thanks for the PR regardless @luposlip

PR #596, opened 2020-05-16
Fix typo 'chnage' to 'change' in dominoes-30k.md

Blevs:
As a side note, the github description for the project states 'building uer interfaces' instead of `building user interfaces'.


superstructor:
Although this was submitted first I didn't notice it until I had already merged #601 which appears to be a duplicate.

Thanks all the same for the fix! @Blevs 😄 Please keep up PRs for any more issues you spot. Also GitHub description fixed.

PR #600, opened 2020-05-18
missing closing bracket in dispatch example

jdkealy:
the example dispatch function [here](http://day8.github.io/re-frame/codox/re-frame.core.html#var-dispatch) is missing a closing bracket


mike-thompson-day8:
Thanks!

Issue #614, opened 2020-06-24
v1.0.0-rc2 doesn't work on Webworker


2knu:
I'm using re-frame in webworker (primary to use pneumatic-tubes for dispatching events from serverside over an websocket to webworker). With the newest v1.0.0-rc2 of re-frame it doesn't work anymore, with v0.12.0 it worked.

## Error message
 ```
Uncaught ReferenceError: window is not defined
    at re_frame$interop$on_load (interop.cljs:9)
    at settings.cljc:13
 ```

## Possible Fix
Using js/self instead of js/window (in interop.cljs) would fix this issue (tested in webworker and main thread). In main thread it gives the window object, in webworker it gives the worker object.

 ```
(defn on-load
      [listener]
      (events/listen js/self "load" listener))
 ```

superstructor:
Should be fixed by #615

PR #615, opened 2020-06-24
Fixing running re-frame on webworker

2knu:
Fixes issue #614


superstructor:
Thanks @2knu 😄 👍 

I also double checked that `self` is safe in all browsers and [according to MDN it is](https://developer.mozilla.org/en-US/docs/Web/API/Window/self).

PR #620, opened 2020-06-30
Fix image path in event-handling-infographic.md

lucasweng:
Thanks for the infographics! I came across this while reading the doc.


mike-thompson-day8:
Thanks!

PR #622, opened 2020-07-01
Fix 404 error and point now in new docs URL

PrestanceDesign:
Which one it's better ? 
I guess it's better now to point on the fresh new docs instead on the github file https://github.com/day8/re-frame/blob/master/docs/subscriptions.md


mike-thompson-day8:
Thanks!

PR #624, opened 2020-07-02
Removing an extra div

Panthevm:
:<> [Reagent Fragment ](https://github.com/reagent-project/reagent/blob/master/doc/ReactFeatures.md)
### Example 1

```
[:div
  [:span "foo"]
  [:span "bar"]] 
```

=>
```
<body>
   <div>
      <span>"foo"</span>
      <span>"bar"</span>
  </div>
</body>
```
### Example 2

```
[:<>
  [:span "foo"]
  [:span "bar"]] 
```

=>
```
<body>
    <span>"foo"</span>
    <span>"bar"</span>
</body>
```

### DIFF
--- a/examples/todomvc/src/todomvc/views.cljs
+++ b/examples/todomvc/src/todomvc/views.cljs
@@ -99,7 +99,7 @@
 
 (defn todo-app
   []
-  [:div
+  [:<>
    [:section#todoapp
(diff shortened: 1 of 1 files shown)


mike-thompson-day8:
Thanks!

Issue #627, opened 2020-07-05
Improve exception debugging by removing catch and rethrow from re-frame's router


mike-thompson-day8:
I have verbally received the following suggestion, which is aimed at making it easier to debug event handlers when they throw exceptions  ...

Currently, when an exception is thrown in an event handler, it is caught by the router, which does some cleanup, and then rethrows.  This catch and rethrow business is a nuisance - it adds a little sand into the gears. And it probably isn't actually needed - can it be removed?

Why? Chrome's devtools will break on uncaught exceptions, which is good. It will give you the call stack and locals at the throw point. ***But*** when the exception is rethrown (as is the case in re-frame event handlers) it is the rethrow point which is reported - what we'd like is for devtools to stop at the original throw point (the event handler) and not the rethrow point (deep in the guts of re-frame's router)

mike-thompson-day8:
Unfortunately, the PR Isaac and I developed to mitigate this issue makes the router code ugly and a bit fragile.  I am not keen to merge it.  We'll have to think of a better way to achieve this or abandon this requirement.

PR #628, opened 2020-07-06
Improve event exception debugging

superstructor:
Removes catch and rethrow from router re #627


superstructor:
Closing as this is not a basis for a solution (unsolvable cases). #627 remains open to track the original issue.

PR #631, opened 2020-07-08
Fix Basics -> Code -> Domino 2 code example

jacobemcken:
Database (or `db`) seems to be the preferred way to refer to the state other places in the documentation, but since the rest of the example revolved around "state" I kept that.


mike-thompson-day8:
Thanks!  
Yes, right choice, there were too many `db` in that exact part of the docs, so I (recently) switched to `state` for clarity.

PR #634, opened 2020-07-10
Feature Request: Syntactic sugar for trivial reg-sub declarations.

bsboiko:
### Syntactic sugar for direct access to input signals in registering subscriptions

Working with re-frame, I have developed a number of patterns and noticed some areas where there is a non-significant amount of boilerplate that I tend to include in this work. In specific, a vast quantity of subscriptions have a single, trivial purpose, to get either a positional or associative value from some input signal.

```clj
(reg-sub
  :query-id
  (fn [db _]
    (:foo db)))

;; or

(reg-sub
  :other-query-id
  (fn [db _]
    (first db)))
```

In this case, the computation function covers a very small operation, and there is a small overhead in writing (and slightly for reading) an anonymous function to get this value. We can of course reduce this to:

```clj
(reg-sub
  :query-id
  #(:foo %))
```

While more concise, this is still not as good as we can get, and to be honest, it doesn't visually look as nice. Ideally, I would like to be able to simple use a keyword (as it implements `IFn`) or really any function right away:
```clj
(reg-sub
  :query-id
  :foo)
```

This form is very ideal, since it clearly represents the computational purpose of the subscription, and it is very little boilerplate. We are getting the value at `:foo` and that is all we need to know. However, there is a slight concern here as the computation function is a 2-arity (I suppose technically 3?) which can introduce potential danger here. If for any reason, `:foo` is not present in our input signal, the query vector as the second position will become the default, and instead of returning `nil` as we would want, we would get `[:query-id]`. This result is misleading and inaccurate.

For a while, I have been using a wrapper for `reg-sub` that will take the last argument and reduce it to a 1-arity function that only consumes the input signal and replaces that as the computation function for me. The implementation of the few functions involved is below:
```clj
(defn sub-get
  [f]
  (fn [data _]
    (f data)))

(defn my-reg-sub
  [& args]
  (let [args (conj (vec (butlast args))
                   (sub-get (last args)))]
    (apply reg-sub args)))

(my-reg-sub
  :query-id
  :foo)
```

This has worked really well, and it lets me write subscriptions for such simple use cases in a much more compact way and ensure my data will be correct. And I still had the original implementation `reg-sub` any time I needed all of its power. As I continued to jump between other re-frame projects, I grew a bit fatigued having to carry this small helper function between them, and I got a bit frustrated with the appearance. I wondered if it would be feasible to implement these changes in re-frame directly and add it to `reg-sub` so I could have that power natively in re-frame.

The first concern was finding a clear way to provide a distinction between using a regular 2-arity computation function, and this special, input signal only, function. Following in the syntactic sugar provided by the signal functions, I added an extra optional keyword that can indicate the computation function, or output path:
```clj
(reg-sub
  :query-id
  :-> :foo)
```

The presence of `:->` creates a higher order function following the implementation of `sub-get` above, and let's me gain direct access to the input signal. I chose this symbol as it pairs nicely with the `:<-` that comes with signal functions. In fact, the presence of this existing syntactic sugar was a huge piece of inspiration for this train of thought. They signify inputs, and my sugar here shows the flow of data to the eventual output.
```clj
(reg-sub
  :query-id
  :<- [:another-query-id]
  :-> :foo)
```

Once again, this can be any function or item that implements `IFn`, and this opened me to a world of writing composable, functional subscriptions. Here are some examples:
```clj
(reg-sub
  :query-id
  :<- [:another-query-id]
  :-> #{:bar})             ;; <---- A concise way to check for equality for my input value. Could also be #(= :bar %)

(req-sub
  :sorted-data
  :<- [:unsorted-data]
  :-> (partial sort-by <)) ;; <---- A concise way to denote consumption of an input sub and sorting its data

;; app-db {:a 1 :b 2}
(reg-sub
  :a-sub
  :-> :a)

(reg-sub
  :b-sub
  :-> :b)

(reg-sub
  :a-b-sub
  :<- [:a-sub]
  :<- [:b-sub]
  :-> (partial zipmap [:a :b]))

;; Before would have something like (fn [[a b] _] {:a a :b b})
;; We get to move away from this declarative style to a functional approach
```

In short, this direct access to only the input value allows me to write trivial subs in shorter code and with greater readability and clarity, and it also allows me to write more succinct, and functional Clojure code. And I am able to do it directly from a version of re-frame without any wrapper functions. All of those are wins in my view.

I am interested in hearing thoughts, questions, concerns on this and potentially even viability for use within re-frame. I do know that this can be done externally with very little overhead, but the notion of having it done right here and sharing it with other re-frame developers is exciting to me. I implemented the functionality for this in `reg-sub`, and I added extensive testing for it as well.

### Extra sugar

I added another bit of syntactic sugar, which I am not as enamored by, but I implemented it anyway as a thought experiment and debate point. Also caffeine is a wonderful thing.

In addition to the pattern of directly operating on the input signal, my other large use case for`reg-sub` is to take the input signal, and some extra, optional values passed into the query vector. I have used the query id contained in the first position of this maybe a handful of times in my re-frame time. Having access to it is nice, but I almost always omit it. In my whiteboarding, I realized I could get value by being able to optionally have a computation function that expected parameters in the form of `[input-signals & optional-query-values]` instead of `[input-signals [query-id & optional-query-values]]`. This would again be able to help me remove a number of trivial subscriptions, checking for equality, or combining data:
```clj
(reg-sub
  :query-id
  :<- [:another-query-id]
  (fn [data [_ q]]
    (= data q)))

(req-sub
  :other-query-id
  :<- [:another-another-query-id]
  (fn [data [_ q]]
    [data q]))
```

This is a non-insignificant amount of boilerplate just to be able to check for equality. Instead, I added another piece of sugar that allows me to get a multi-arity function that explodes the query optional values right next to the input signals using the marker `:=>`. The above rewritten this way would be:
```clj
(reg-sub
  :query-id
  :<- [:another-query-id]
  :=> =)      ;; <---- This is the same as (fn [data q] (= data q))

(req-sub
  :other-query-id
  :<- [:another-another-query-id]
  :=> vector)    ;; <---- Or (fn [data q] [data q])
```

I will admit that this second piece is a bit more of a stretch, but I do have a number of subscription patterns I write that could benefit greatly from access to this data on the same level. I also implemented this, and tested this. I would like to have a conversation about this as well, and see if it has any future in re-frame.

Both of these would significantly remove boilerplate code from my subscriptions and allow me to spend my time on more interesting uses of subscriptions.


superstructor:
Agreed. It will be in the next release. Thanks @bsboiko 👍🏻 😄

PR #637, opened 2020-07-18
Add missing character and clarify meaning in dominoes-30k

alaq:
Just a small typo I noticed when reading the documentation!


mike-thompson-day8:
Thanks!

PR #650, opened 2020-10-17
fix markdown formatting

phronmophobic:
The markdown rendering of the list looks right on github, but appears as a jumble on the actual re-frame documentation site, https://day8.github.io/re-frame/reusable-components/


mike-thompson-day8:
Thanks!

PR #653, opened 2020-10-26
Doc list fixes

phronmophobic:
Some lists aren't being displayed correctly on the re-frame docs website. I actually got the doc generation set up locally to test these fixes this time!


mike-thompson-day8:
Thanks!

PR #654, opened 2020-10-26
Fixing doc generation command examples.

phronmophobic:
I'm not sure what the slashes in the `ns-to-markdown` command are the wrong direction. Is that the windows command?

I also think the windows and linux commands were swapped for the docker commands.

The docs didn't specify which directory to run the docker commands from. I mistakenly thought it was supposed to be run from `docs/` given the previous commands.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -28,16 +28,19 @@ cd re-frame
 Then build the API documentation:
 ```sh
 cd docs/
-clojure -m ns-to-markdown ..\src\re_frame\core.cljc > api-re-frame.core.md 
+clojure -m ns-to-markdown ../src/re_frame/core.cljc > api-re-frame.core.md
 ```
 
-Then, if using PowerShell on Windows:
+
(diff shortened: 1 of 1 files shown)


mike-thompson-day8:
I'm on windows and my experience is .... 

image

phronmophobic:
Ah, ok. I'm on a mac and the `${PWD}` version is also the version that worked for me. I know what `${PWD}` means, but I don't know what `%cd%` is supposed to do so I assumed it was some windows things (I'm very unfamiliar with windows).

mike-thompson-day8:
I have updated the README with your ideas.
https://github.com/day8/re-frame/commit/1a257100a1778566fd600cf9733a091e1487ade5

Thanks!

mike-thompson-day8:
And one more:  https://github.com/day8/re-frame/commit/e7137051742595c37f673f5be68531f40ac18a84

PR #659, opened 2020-11-20
[tutorial todomvc] remove duplicate css option: boarder

MokkeMeguru:
I think the duplicates option is incorrect.
(this pr is a too small change.  thanks for your review  )


superstructor:
Thanks @MokkeMeguru 👍

PR #660, opened 2020-12-06
docs: fix improperly indented code block

msladecek:
Code block in docstring isn't indented properly so it isn't rendering correctly in web docs http://day8.github.io/re-frame/api-re-frame.core/#subscriptions.

image


superstructor:
Thanks @msladecek 👍

PR #661, opened 2020-12-08
Fix typos: it's → its

abigsmall:
First ran into the docs typo when reading through the FAQ, and then decided to check all other instances of `it's` in the repo to make sure there weren’t any other cases in which the contraction (it is/it has/it was) was being used instead of the possessive.


mike-thompson-day8:
Thanks!

PR #668, opened 2021-01-10
Bring :timer snippet in sync with simple/core.cljs

rogererens:
The line of code appears redundant as it is neither explained with an in-line comment following it nor in the notes below the snippet.
The line of code is also not present in examples/simple/src/simple/core.cljs


mike-thompson-day8:
Thanks!

PR #730, opened 2021-05-21
Simplified calls to `apply`.

green-coder:
I wasn't able to run the tests, but it should work.

### DIFF
--- a/src/re_frame/core.cljc
+++ b/src/re_frame/core.cljc
@@ -338,7 +338,7 @@
   "
   {:api-docs/heading "Subscriptions"}
   [query-id & args]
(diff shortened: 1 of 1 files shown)


superstructor:
Nice simplification @green-coder. Thanks 👍🏻 😄 

Why were you not able to run the tests ? Is there something that needs to be fixed ? 

Anyway, the tests in this case wouldn't have mattered as those fns (`on-changes`, `reg-sub`) are tested directly via their respective ns (`re-frame.std-interceptors`, `re-frame.subs`) and not via the wrappers that have been changed in `re-frame.core`.

So I've verified via a REPL that this should work as intended.

PR #731, opened 2021-05-21
(last pair) -> (second pair)

green-coder:
`second` uses less resources and looks more appropriate for a fixed-size vector.

### DIFF
--- a/src/re_frame/subs.cljc
+++ b/src/re_frame/subs.cljc
@@ -176,7 +176,7 @@
                          ;; multiple sugar pairs
                          (let [pairs   (partition 2 input-args)
                                markers (map first pairs)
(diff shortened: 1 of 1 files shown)


green-coder:
On a second thought, maybe [`key`](https://clojuredocs.org/clojure.core/key) and [`val`](https://clojuredocs.org/clojure.core/val) could be better - I will take a look later today.

superstructor:
Nice find 👍🏻 😄 @green-coder Since `last` uses `recur` and `second` uses `INext` I agree its probably more efficient.

I'm also confident this will not change behavior since `partition` (line 177) without a `pad` will indeed always return fixed-size pairs of `cljs.core/LazySeq` (not vector tho, but no matter).

PR #738, opened 2021-08-31
Add support for multithreading dynamic appdb

werenall:
This commit implements the idea explored by Techascent in [this blogpost](https://techascent.com/blog/isomorphic-rendering.html).

It accomplishes it by changing appdb to be a dynamic variable.
Then, when needed, one could serve an html with this snippet:
```clojure
(defn- handle-route*
 [app-db]
 (let [db-id (util/uuid)
       result (with-bindings {#'re-frame.db/app-db-id db-id}
                (swap! rf.db/db-atoms* assoc db-id app-db)
                (gen-html app-db))]
   (re-frame.db/clear-app-db db-id)
   result))
```

It makes sure that the cljs behavior is unchanged and only adds clj support.

### DIFF
--- a/.idea/codeStyleSettings.xml
+++ b/.idea/codeStyleSettings.xml
@@ -1,21 +1,21 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
-  <component name="ProjectCodeStyleSettingsManager">
-    <option name="PER_PROJECT_SETTINGS">
-      <value>
-        <ClojureCodeStyleSettings>{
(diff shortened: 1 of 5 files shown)


coltnz:
Hi @werenall, thank you for the contribution. 

We've read through your patch and can see it would provide some value, but at this point we're (Day8) still focussed on the more ambitious (and overlapping) ideas of 'multi-instance re-frame'. See [EP2](https://github.com/day8/re-frame/blob/master/docs/EPs/002-ReframeInstances.md]) for more on this.

PR #744, opened 2021-10-10
Fix for documentation regarding documentation docker images

nikkez:
It seems that the commands that start the docker images for the documentation are in wrong order. The command that is mentioned for Windows environments works for Linux environments but not the other way around.


superstructor:
Thanks @nikkez 👍🏻 😄

PR #748, opened 2021-12-21
Some doc fixes

eval:
This fixes: 
- codeblocks for [the docs of reg-sub](https://day8.github.io/re-frame/api-re-frame.core/#reg-sub).  
- the list[^1] in that description should contain 3 items and not be two separate lists.  
- some spelling errors 

[^1]: Starting at `The computation function is always the last argument supplied and has three ways to be called. ...`


superstructor:
Thanks @eval 👍🏻 😄

PR #749, opened 2021-12-27
Use fx vectors in reg-event-fx in docs

NoahTheDuke:
Given that the new recommended behavior is to use `:fx` in `reg-event-fx` return maps, it seems prudent to use this format in all of the documentation as well. I've done my best to cover all of the usages and update the language to correctly match the implementation.

I also fixed a couple mistakes around missing or extra parentheses and brackets in code blocks.

### DIFF
--- a/docs/EffectfulHandlers.md
+++ b/docs/EffectfulHandlers.md
@@ -228,7 +228,7 @@ Here it is re-written so as to be pure:
    :my-event
    (fn [{:keys [db]} [_ a]]                ;; <2>
       {:db  (assoc db :flag true)          ;; <3>
-       :dispatch [:do-something-else 3]}))
+       :fx [[:dispatch [:do-something-else 3]]]}))
 ```
 
 Notes: <br>
@@ -239,7 +239,7 @@ it is a map which contains a `:db` key. <br>
 *<3>* The handler is returning a data structure (map) which describes two side-effects:
 
   - a change to application state, via the `:db` key
-  - a further event, via the `:dispatch` key
+  - a further event, via the `:fx` key calling the `:dispatch` effect handler
 
 Above, the impure handler **did** a `dispatch` side-effect, while the pure handler **described**
 a `dispatch` side-effect.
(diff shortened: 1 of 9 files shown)


docs/EffectfulHandlers.md
```diff
@@ -239,7 +239,7 @@ it is a map which contains a `:db` key. <br>
 *<3>* The handler is returning a data structure (map) which describes two side-effects:
 
   - a change to application state, via the `:db` key
-  - a further event, via the `:dispatch` key
+  - a further event, via the `:fx` key calling the `:dispatch` effect handler
```
p-himik, line 242: I'd say "containing" instead of "calling". There's no calling here, yet.

docs/EffectfulHandlers.md
```diff
@@ -337,7 +337,7 @@ data from arguments.
 `-db` handlers take __one__ coeffect called `db`, and they return only __one__  effect (db again).
 
 Whereas `-fx` handlers take potentially __many__ coeffects (a map of them) and they return
-potentially __many__ effects (a map of them). So, One vs Many.
+potentially __many__ effects (a map of a vector of them). So, One vs Many.
```
p-himik, line 340: `:fx` is just yet another effect, albeit a meta one in this case. `-fx` handlers still do return a map of effects - this change is wrong.

p-himik:
A comment on the PR overall - I'm not sure there's a good reason to replace a single effect with `:fx`. It makes the code longer and it brings no real value. It _could_  potentially bring more value if one wrote everything in a different way to allow easy handler composition, but it's not what this documentation is about.
So if there are at least two effects (besides `:db` since its order is well defined), I would probably replace it with `:fx`. But if there's just one such effect, personally I wouldn't touch it. But of course I'm not a re-frame maintainer, so that's just my own 5 c.

NoahTheDuke:
@p-himik Thank you for the input! I think that's a solid read of the situation. It seemed to me that the intention of the `:fx` handler was to move away from "top-level" `reg-event-fx` keys and to use `:fx` as the primary means of performing non-db side effects. If that's _not_ the case, then maybe the [Order of Effects section](https://day8.github.io/re-frame/Effects/#order-of-effects) should be changed so that it doesn't call using `:fx` like this "best practice"?

I'll wait to make any further changes until one of the maintainers speaks up.

mike-thompson-day8:
I appreciate the work, but I'm a little cautious about doing this.  I'm still dwelling on it. 

With docs, there is always a tension between docs being "complete" and "accurate" and docs being helpful (by leaving out complexity).  API/reference docs should always be the former. Whereas tutorial introductions should be the latter. because they should explain concepts as simply as possible, even if they are slightly inaccurate.  

I think the docs need a "best practice" document which is aimed at the intermediate level.  THAT would be the place to insist on use of `:fx` etc.

NoahTheDuke:
Thanks for the detailed response! Everything you mention here is good and I appreciate the delicate balance between precision and ease of understanding. I'll leave this PR open but feel free to close it as you choose.

NoahTheDuke:
After working more with re-frame, I think this PR is not ideal. Closing for now.

Issue #740, opened 2021-09-07
[Enhancement]: Warn on `subscribe` being used outside of a reactive context


p-himik:
### What do you suggest?

People often make the mistake of using `subscribe` in a JS event handler or in a re-frame event handler, probably in other places. To know better, one have to carefully read the documentation, which people often don't do.

It would be better to warn them about `subscribe` being used incorrectly with a link to the right section of the documentation.

qleguennec:
On a side note, I've not been entirely convinced that this pattern is effectively an issue (according to the docs, it "might create a memory leak", can't it be more precise?). I'm a dev in an fairly old re-frame project and this pattern is used extensively, and it seems to work just fine. 

Is the purpose of the warning only to warn that the subscription won't be reactive, or will there be something else added to it?

p-himik:
Subscriptions are cached and the cached is pruned from an unused subscription only when the last watcher of that subscription is removed.
When a sub is never used in a reactive context, there are no watchers at all. It's still added to the cache but it's never removed - creating a memory leak. The leak is only potential because the same sub could also be used in a reactive context.

dannyfreeman:
Another thought would be to detect if a call to subscribe is happening in a reactive context, and if it is not bypass caching the subscription in the first place. That would avoid the memory leak.

PR #752, opened 2022-02-07
Warn user when calling subscribe outside of a reactive context

dannyfreeman:
I saw the help wanted tag on issue #740 and decided to open a PR for it

This will log warnings to the console every time a call to subscribe is made in a place it shouldn't be called, like event handlers.

Here is what the warning looks like in firefox:
image

Right now it is printed as a warning. It may be more useful to print it as an error so the user can see the stack trace and find what code triggered the message.

### DIFF
--- a/src/re_frame/interop.clj
+++ b/src/re_frame/interop.clj
@@ -91,3 +91,7 @@
   "Doesn't make sense in a Clojure context currently."
   [reactive-val]
   "rx-clj")
+
+(defn reactive?
+  []
+  true)
--- a/src/re_frame/interop.cljs
+++ b/src/re_frame/interop.cljs
@@ -72,3 +72,7 @@
            reagent.ratom/Track "tr"
            "other")
          (hash reactive-val))))
+
+(defn reactive?
+  []
+  (reagent.ratom/reactive?))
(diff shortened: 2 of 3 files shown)


dannyfreeman:
I put this in the issue, but wanted to drop it here too. It may be worth figuring out how to change re-frame such that this warning is not necessary.

I made this branch a while back to explore what seems to be a reasonable solution to the problem:
https://github.com/dannyfreeman/re-frame/compare/master...dannyfreeman:bypass-subscription-cache-outside-reactive-context#

In short, when the current call to subscribe is not happening in a reactive context and a subscription is not already cached, then bypass the subscription cache. Just return a new reaction. If the subscription was already cached then any code can grab and derefence that cached subscription (which is now only cached in reactive contexts, so something will eventually clean it up).

superstructor:
Thanks @dannyfreeman 👍🏻 Code is good quality. 

Please raise a new issue for the solution described in https://github.com/day8/re-frame/pull/752#issuecomment-1032139326 as my understanding is it would be best separated as a distinct future change to this even if it'll replace this warning.

dannyfreeman:
Thank you @superstructor 
I followed up with a new ticket https://github.com/day8/re-frame/issues/753 and a draft PR to go along with it https://github.com/day8/re-frame/pull/754

Issue #753, opened 2022-03-06
[Enhancement]: Allow subscribe to be safely called from anywhere


dannyfreeman:
### What do you suggest?

This ticket is a followup from PR https://github.com/day8/re-frame/pull/752#issue-1125508731

### Enhancement Request
I would like to be able to call `subscribe` anywhere in my re-frame app without the risk of leaking memory in the subscription cache. Specifically, I would like to be able to call subscriptions from inside events. There is a lot of useful business logic performed in subscriptions that can be re-used in event handlers and other non-reactive contexts.

### Current workarounds
If the event is close to the view (I.E. dispatched directly in a click handler or something like that), I use the subscription inside view and pass the value to the event when dispatching. This is usually really easy when I also need the subscription value in the view itself

Sometimes my event is a little further down the dispatch chain, or dispatched by many views that don't need to subscribe to the same data themselves. In this situation I extract subscription handlers into standalone functions with a signature of `db & args` and call that function from the subscription handler and the event. This works fine for simple layer-2 subscriptions. For layer 3 subscriptions I have to re-create the input signal graph manually. This means extracting regular functions out of many subscription handlers and composing the results, a process that can be messy and prone to error.

### Proposed Solution
When calling subscribe and a reaction is not present in the subscription cache, check that the current procedure is being called in a reactive process before caching. If it is not reactive, bypass caching the reaction. I have a patch in the works for this.

This could result in a lot of cache misses when calling subscriptions from events, but the performance penalty is nearly identical to workaround I most commonly employ, which is extracting subscription handlers into functions and invoking those functions every time the event handler executes. If the subscription is cached then the event handler will be able to use the cached reaction, which is a nice improvement over the current state of affairs.

#### Other possible solutions
Some other solutions I have entertained: a special subscribe function that is only intended to be used outside reagent components, or a co-effect for injecting subscription values in events.
Having different semantics for different contexts could have it's own advantages, like keeping the existing "don't sub outside a component" warnings where people really shouldn't be subscribing, like an `on-click` callback. The proposed solution above wouldn't be able to differentiate between that situation and subscribing in an event.

PR #754, opened 2022-03-06
Allow subscribe to be safely called from any context

dannyfreeman:
The `warn-when-not-reactive` warnings are no longer valid if we bypass
the subscription cache when not in a reagent component, so they've been
removed.

This should address issue #753

Still needs some testing, and probably some documentation cleanup


mbertheau:
I love this. It seems to be an easy way – API-wise – to make values from the subscription calculation available in event and effect handlers.

What seems a bit problematic, or at least questionable (in the literal sense), is: what `db` state is used? In event handlers and possibly effect handlers you work with data passed into the function. `subscribe` uses a potentially different version of the app-db.

It seems to me that the API that one would want in that situation is `(get-subscription-value db [:subscription parameters])`, i.e. you pass in the db that should be used.

mbertheau:
> There is only one db state in existence at a time, it will not be changed out from under running code in a single threaded environment. If you call a subscription in an event then the db value will be the same as the db being used in the event.

Yeah. What I mean is that you might want to get the result of a calculation captured in a subscription based on the db that you're currently building a new version of in the event handler.

Some pseudo code demonstrating the mechanics I have in mind, not an actual use case:
```clojure
(reg-sub :count-foo [db] (-> db :foos count))

(reg-event-handler :button-clicked [db new-foo]
  (let [db (-> db (update :foos conj new-foo)]
    (if (> 10 (get-sub-value db :count-foo))
      (assoc db :big? true)
      db)))
``` 



> 
> If we want something like `get-subscription-value` then a subscription can just be extracted into a function.

Well, not "just" because you have do extract all input subscriptions as well and compose everything together, and that might not be trivial, and/or impose significant boilerplate code when all subscriptions are written in a way that allows for that.

> Writing a generic utility for this would involve re-implementing subscriptions without de-referencing the global app-db. I've done something similar with re-frame to allow atoms other than app-db to be used. It was a lot of work. That is not something i want to waste my time on again because the re-frame maintainers don't seem interested in large patches like that.

Sounds like a challenge! :)

PR #755, opened 2022-03-31
Update External-Resources.md

tonyxrandall:
added an article


mike-thompson-day8:
Is this article intended to be general information? 
Because I can't see any direct mention of re-frame? 
Doesn't seem sufficiently relevant. 
I'll close, but please push back if I've missed something.

PR #764, opened 2022-09-17
Fix example reference in reg-sub docstring

joshuamzm:
Reading through the `reg-sub` docstring (https://day8.github.io/re-frame/api-re-frame.core/#reg-sub), I found confusing a reference to a function `input-fn` not previously mentioned in the document. From context, I understood it referred to the signal function used as the second argument of `reg-sub`.


superstructor:
Thanks @joshuamzm Good spotting.

PR #772, opened 2023-01-25
Use correct evaluation function in docs example

osvalds:
I noticed a tiny issue in the docs where `if` should have been used instead of `when`.


mike-thompson-day8:
Thanks!

PR #773, opened 2023-02-05
Update examples of applications built with re-frame

wcalderipe:
This PR was based on the suggestions given by @p-himik in the #re-frame channel of clojurians Slack workspace.

See also https://clojurians.slack.com/archives/C073DKH9P/p1674292326385029


mike-thompson-day8:
Thanks!

PR #774, opened 2023-02-09
Add Reflet suite to External-Resources.md

zalky:
**Reflet**: Fast multi-model db, mix graph/non-graph data, graph queries/mutations, hierarchical FSMs, JS/DOM interop utils, component APIs, novel visual debugging

See here for more details: https://github.com/zalky/reflet


mike-thompson-day8:
Thanks!

PR #789, opened 2023-07-18
minor refactor of simple example to align with docs

### LINKED ISSUE #758: [Bug] [Docs]: Simple app example docs out of sync with code in repo
### What happened?

I recently started looking into re-frame (looks great, I wish the JS community was as wise when it comes to architecture) and read through the docs. The follow-along with simple app docs https://day8.github.io/re-frame/dominoes-live/ was out of sync with the code (which the docs link to) https://github.com/day8/re-frame/blob/master/examples/simple/src/simple/core.cljs. Most notably the views differed, but there are other minor differences too (e.g. the name of the mount-ui/render function).

(If you want I could have a go at fixing it, but as mentioned I'm very much a re-frame novice. I would need to know which of the two is the correct version).

### Version

1.2.0

### What runtimes are you seeing the problem on?

Google Chrome

### Relevant console output

_No response_

ag91:
I found myself looking at how the example in the docs was written (klipse is cool indeed) and I saw the issue,
so I thought one less on the issue tracker ;)
(Thanks for maintaining re-frame!)

Fixes #758

### DIFF
--- a/examples/simple/src/simple/core.cljs
+++ b/examples/simple/src/simple/core.cljs
@@ -25,7 +25,7 @@
  :initialize                 ;; usage:  (dispatch [:initialize])
  (fn [_ _]                   ;; the two parameters are not important here, so use _
    {:time (js/Date.)         ;; What it returns becomes the new application state
-    :time-color "#f88"}))    ;; so the application state will initially be a map with two keys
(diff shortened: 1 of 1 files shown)


kimo-k:
46ecc4ad
00891079

Issue #809, opened 2024-11-15
[Bug]: Non-daemon executor thread in Clojure prevents clean JVM shutdown


oliyh:
### What happened?

Hi,

Firstly thank you for the great library, it transformed the way I thought about building UIs and it's so much more elegant than redux and so much cleaner than React hooks.

I have a library called re-graph which uses re-frame mostly for state management of websocket lifecycle. It runs in both cljs and clj, and we have uncovered an issue with the clj implementation which is preventing clean shutdown of the JVM because of an executor thread which is left running.

The full conversation is here: https://github.com/oliyh/re-graph/issues/75

@lowecg has produced some code which works around the issue, but I think this would be best incorporated into re-frame (or maybe re-frame-test) and I'd be grateful to know what you think.

Thanks for your time

### Version

1.2.0

### What runtimes are you seeing the problem on?

JVM (CLJC)

lowecg:
I've created a minimal project demonstrating the issue: https://github.com/lowecg/re-frame-issue-809

I'll issue a PR shortly

lowecg:
@kimo-k / @oliyh  Thanks for you time on this. PR raised - I hope it meets expectation :-) Let me know if you need anything else.

kimo-k:
My takeaways so far:

- calling `dispatch` actually pushes an event (vector) onto a [queue](https://github.com/day8/re-frame/blob/f654f42430ac0bad9ad43688ebc9d712fb26b4e6/src/re_frame/router.cljc#L222)
- when re-frame pulls an event from the queue (to call the event handler), that's called a [tick](https://github.com/day8/re-frame/blob/f654f42430ac0bad9ad43688ebc9d712fb26b4e6/src/re_frame/router.cljc#L139)
- this queue [defers to the host environment](https://github.com/day8/re-frame/blob/f654f42430ac0bad9ad43688ebc9d712fb26b4e6/src/re_frame/router.cljc#L31), leaving time for stuff like draw calls & (non-re-frame) event loops
- when running re-frame on the JVM, ticks are [handled by an executor](https://github.com/day8/re-frame/blob/f654f42430ac0bad9ad43688ebc9d712fb26b4e6/src/re_frame/interop.clj#L31).
- the executor needs an explicit shutdown call. Otherwise, [if you try to exit the JVM process, it fails to exit](https://github.com/lowecg/re-frame-issue-809).
[...]

PR #810, opened 2024-11-17
Add reset-executor! and shutdown-executor! to re-frame.interop

lowecg:
For https://github.com/day8/re-frame/issues/809

**Executor Service Lifecycle Management**
Added two functions to manage the executor service lifecycle in re-frame/interop:

reset-executor! - Creates a fresh single-threaded executor when needed
shutdown-executor! - Cleanly terminates the executor with configurable timeout

**Why?**
These functions enable proper cleanup in short-lived processes like AWS Lambda, CLI runs and tests. Without explicit shutdown, the non-daemon executor threads prevent the JVM from exiting cleanly.

Adding the following to a test ensures clean shutdown:

```clojure
(ns my-ns
  (:require
    [clojure.test :refer :all]
    [re-frame.interop :as interop]
    ...))

(use-fixtures :once
              (fn [f]
                (f)  ; runs all tests
                (interop/shutdown-executor! 100)))
```

### DIFF
--- a/src/re_frame/interop.clj
+++ b/src/re_frame/interop.clj
@@ -1,5 +1,5 @@
 (ns re-frame.interop
-  (:import [java.util.concurrent Executor Executors]))
+  (:import [java.util.concurrent Executor ExecutorService Executors TimeUnit]))
 
 ;; The purpose of this file is to provide JVM-runnable implementations of the
 ;; CLJS equivalents in interop.cljs.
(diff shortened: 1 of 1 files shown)


kimo-k:
some discussion - https://clojurians.slack.com/archives/C073DKH9P/p1733141055817809

lowecg:
I've removed reset-executor! from the PR. Lambda functions work fine without interacting with the scheduler

p-himik:
@lowecg Have you seen the discussion on Slack? Do you have any thoughts on getting rid of the executor completely and running all re-frame stuff in CLJ on the main thread?

lowecg:
Hi Eugene,

I have read the discussion - thank you all for your inputs!

The point about not exposing an API to the executor was compelling, and I
fully agree there.

In terms of removing the executor, that makes sense given that it appears
not to be emulating CLJS behaviour, in fact it's changing it. Not having
the executor would prevent the need for a workaround, and would prevent
users walking into the issue in the first place.

Would you like me to change the PR to remove the executor and change
next-tick as per the discussion?

Cheers,

Chris

p-himik:
@lowecg Ah, sorry for the noise - I removed my message pretty much immediately after I realized that I've posted it on the PR and not on the issue page.

As for what to do - that would be up to @kimo-k or someone else from Day8. But were it my own PR, I would definitely go the route of removing the executor, the thread for `next-tick`, and the associated locking that was added only to make the thread behave properly.

kimo-k:
Regarding locking: #469 #471

lowecg:
@p-himik @kimo-k Thank you for your continued input on this.  I've removed the executor and locking added for #471. The change fixes the repo cases I'd raised for #809. I've left the test from #469 since it shows the simplified next-tick is working as intended in this regard.
 
Here's a short summary for the change, I hope it sufficiently captures your thread on Slack:
 
The single-threaded executor in the JVM implementation of re-frame is being removed to simplify the code and resolve concurrency issues. Originally added to mimic JavaScript’s single-threaded behaviour, it introduced unnecessary complexity and inconsistent behaviour, such as race condition,  unpredictable event handling orders and process hang when running tests from the CLI.

Benefits to removing the executor:

- Simpler Code: Eliminates threading and locking overhead.
- Consistent Behaviour: Aligns more closely with JavaScript semantics.
- Improved Reliability: Reduces concurrency issues in tests and development.
- Streamlined Focus: Keeps re-frame’s JVM use focused on testing, avoiding unsupported multithreaded scenarios.

PR #814, opened 2025-04-06
Update dominoes-30k.md

wmealing:
The calling format for the reagent element for onclick had to be in {}.  I'm trying to follow along as I read, so maybe my usage is wrong.


wmealing:
Thank you.

Issue #815, opened 2025-05-07
[Bug]: do-fx does not properly remove flow effects, causing these to run twice


matej-ibis-ai:
### What happened?

A bug happened! It did.

I am developing an app where the user can define multiple queries. In the app-db, I have the `:queries` key whose value is a vector, one element per query. Whenever a query is added, I want associated flows to be registered, so I have an event+effect defined that returns something like:
```edn
{:db new-db
 :fx [[:reg-flow (make-query-flow query-id)]]}
```

Here comes the bug: When I added a log statement to the `make-query-flow` function, I observed it to be printed twice every time the event fired. This indicates the `:reg-flow` effect gets applied twice.

After some debugging using `js/console.trace` and a copy of re-frame code which I kept patching, I figured out that the flow effects are executed once from `re-frame.flow.alpha/do-fx` and once from `re-frame.fx/do-fx`. I have also found what looks like a bug in the first mentioned function, or rather in the `remove-fx` function it relies on; fixing that fixed also the duplicity in the calls. I believe it should be operating on the first element (of vectors or map entries) only:
```clojure
(def remove-fx (partial remove (comp flow-fx-ids first)))
```

Also, in the `do-fx` function, I changed the call of `remove-fx` to collect the result into a vector, not sure if that's necessary:
```clojure
               (-> ctx
                   (update-in [:effects :fx] (comp vec remove-fx))
                   (update :effects (comp (partial into {}) remove-fx)))
```

### Version

1.4.4-SNAPSHOT

### What runtimes are you seeing the problem on?

Mozilla Firefox

### Relevant console output

```shell
17:44:04.865 existing flows: (:snoqb.flows/filtered-items-0 :snoqb.flows/sql-0 :snoqb.flows/bky-sql-0) alpha.cljc:91:13
17:44:04.866 registering flow :snoqb.flows/bky-synced-0 alpha.cljc:92:13
17:44:04.866 console.trace() reg-flow :snoqb.flows/bky-synced-0 stack trace: <anonymous code>:1:145535
    overrideMethod <anonymous code>:1
    cljs$core$IFn$_invoke$arity$1 alpha.cljc:93
    re_frame$flow$alpha$do_effect alpha.cljc:117
    2 core.cljs:4829
    2 cljs.core.js:17910
    sval core.cljs:3511
    cljs$core$ISeqable$_seq$arity$1 core.cljs:3568
    cljs$core$seq core.cljs:1260
[...]

PR #816, opened 2025-05-07
Properly eat flow effects in flows/do-fx

matej-ibis-ai:
This is my attempt at solving #815. I tried to reproduce the issue in a test but I couldn't get the test to fail; instead, I output messages to the console where you can observe they are printed twice instead of once -- on the ~second~ first commit of the feature branch.


kimo-k:
https://github.com/day8/re-frame/issues/815#issuecomment-2875815901

PR #819, opened 2025-07-13
Update interceptor history

KingMob:
The Credit section seemed to say Pedestal created interceptors (that's how I interpret "originally designed" anyway), but they're older than Clojure. 

Both Netty and Tomcat used them at least as far back as the early 00's, and the pattern was documented in at least one book.

I wrote about it a bit more at length in https://clojurians.slack.com/archives/CDG3YQV6Z/p1685076301109489


mike-thompson-day8:
Thanks!

Issue #821, opened 2025-08-13
[Bug]: legacy-sub signal-fn is not backward compatible


RolT:
### What happened?

The signal fn in the `:legacy-sub` mode of the new alpha namespace receives the query **map** instead of the query **vector**. The computation-fn receives the query vector as expected.

```
cljs.user> 
(reg-sub :a 
   (fn [v] (println "Signal argument:" v) (atom nil)) 
   (fn [_ v] (println "Computation argument:" v)))
;; => #object [re_frame$subs$alpha$subs_handler_fn]
cljs.user> (subscribe [:a 1])
Signal argument: {:re-frame/q :a, :re-frame/lifecycle :default, :re-frame/query-v [:a 1]}
Computation argument: [:a 1]
;; => #object [reagent.ratom.Reaction {:val nil}]
```

### Version

1.4.3

### What runtimes are you seeing the problem on?

Google Chrome

### Relevant console output

```shell

```

RolT:
Fixed in https://github.com/day8/re-frame/commit/b916c356f0a6e222d2b4f5f74b0457c7f001fe96

PR #822, opened 2025-08-13
[alpha] fix argument to legacy sub signal fn

RolT:
As described in #821, give the query vector to the input signal function instead of the new query map when using a legacy sub.

Note: this also reverts the trace data, `[:tags :query]` will be the query vector. I'm not sure whether we want this or not.


kimo-k:
Merged in b916c356

